PR #19, opened 2018-02-17
adding extra dominators

### LINKED ISSUE #13: Add options for controlling `svelte dominators` output
* [ ] maximum depth in dominator tree
* [ ] maximum number of rows to include in output, regardless of depth

Each of these involves new fields on the `Dominator` struct in `./opt/opt.rs`.

The `DominatorTree` struct in `analyze/analyze.rs` should make and hold a copy of the `opt::Dominators` options.

Then, the `impl traits::Emit for DominatorTree` should consult the new options and modify its output appropriately.

sendilkumarn:
fixes #13


analyze/analyze.rs
```diff
@@ -196,6 +196,7 @@ pub fn top(items: &mut ir::Items, opts: &opt::Top) -> Result<Box<traits::Emit>,
 
 struct DominatorTree {
     tree: BTreeMap<ir::Id, Vec<ir::Id>>,
+    dominators: opt::Dominators,
```
fitzgen: Nitpick: let's rename this field to `opts`

analyze/analyze.rs
```diff
@@ -212,48 +213,70 @@ impl traits::Emit for DominatorTree {
...
             dominator_tree: &BTreeMap<ir::Id, Vec<ir::Id>>,
             depth: usize,
+            dominators: &opt::Dominators,
             id: ir::Id,
         ) {
             assert_eq!(id == items.meta_root(), depth == 0);
 
-            if depth > 0 {
-                let item = &items[id];
+            if dominators.max_row != 0 && depth != dominators.max_depth {
```
fitzgen: Rather than indenting the whole function inside this `if`, let's:

* add a new parameter, `row: &mut usize`
* add an early return at the start of this function `if *row == opts.max_rows || depth > opts.max_depth { return; }`
* whenever we actually add a row to the table, `*row += 1;`
* finally, when we first call `recursive_add_rows`, pass `&mut 0` for the `rows` parameter

These tweaks should make this functionality a little less invasive

opt/opt.rs
```diff
@@ -125,6 +125,10 @@ pub struct Dominators {
     /// The format the output should be written in.
     #[structopt(short = "f", long = "format", default_value = "text")]
     pub output_format: OutputFormat,
+
+    #[structopt(short = "d", default_value = "10")] pub max_depth: usize,
+
+    #[structopt(short = "r", default_value = "10")] pub max_row: usize,
```
fitzgen: NItpick: `max_rows` instead of `max_row.

Let's also add doc comments to each new field, since this will turn into the CLI help text:

> The maximum depth to print the dominators tree.

> The maximum number of rows, regardless of depth in the tree, to display.

Let's also make these `Option<usize>` rather than providing a default value.

analyze/analyze.rs
```diff
@@ -213,58 +213,70 @@ impl traits::Emit for DominatorTree {
...
+            mut row: &mut usize,
+            opts: &opt::Dominators,
             id: ir::Id,
         ) {
             assert_eq!(id == items.meta_root(), depth == 0);
 
-            if dominators.max_row != 0 && depth != dominators.max_depth {
-                if depth > 0 {
-                    let item = &items[id];
+            if let Some(max_rows) = opts.max_rows {
```
sendilkumarn: Do we have a better syntax here? 

This [RFC](https://github.com/rust-lang/rfcs/issues/929) should be handy

fitzgen: Yeah that would be nice, but it is fine as-is, and better than it was before with the indentation of everything. Thanks again!

PR #20, opened 2018-02-17
adding paths options

### LINKED ISSUE #18: Add ability to specify max path depth and max number of paths
Similar to #13 but for the `svelte paths` subcommand.

Should be able to control how deep we will print paths, and how many total paths we will print.

Check out `opt/opt.rs` for options parsing and `analyze/analyze.rs` for adjusting behavior based on those options.

sendilkumarn:
fixes #18


analyze/analyze.rs
```diff
@@ -276,6 +276,8 @@ pub fn dominators(
 
 struct Paths {
     items: Vec<ir::Id>,
+    max_depth: usize,
+    max_row: usize,
```
fitzgen: Rather than add a copy of each field from the options, and add another field here every time we add an option, we should include a copy of the options here.

```
opts: opt::Paths
```

analyze/analyze.rs
```diff
@@ -289,43 +291,55 @@ impl traits::Emit for Paths {
...
             depth: usize,
+            max_depth: usize,
+            max_row: usize,
             id: ir::Id,
         ) {
-            if seen.contains(&id) || items.meta_root() == id {
-                return;
-            }
+            if max_depth != depth && max_row != 0 {
+                if seen.contains(&id) || items.meta_root() == id {
```
fitzgen: Again, rather than indenting everything, lets fold the new checks into this existing early return.

opt/opt.rs
```diff
@@ -159,6 +159,10 @@ pub struct Paths {
     /// The format the output should be written in.
     #[structopt(short = "f", long = "format", default_value = "text")]
     pub output_format: OutputFormat,
+
+    #[structopt(short = "d", default_value = "10")] pub max_depth: usize,
```
fitzgen: Need doc comments for CLI help.

analyze/analyze.rs
```diff
@@ -323,7 +330,10 @@ impl traits::Emit for Paths {
 
             seen.insert(id);
             for caller in items.predecessors(id) {
-                recursive_callers(items, seen, table, depth + 1, caller);
+                if depth != 0 {
```
fitzgen: This is almost, but not quite what we want. We want to increment the count whenever there is a *new* branch in the path:

```rust
for (i, caller) in items.predecessors(id) {
    if i > 0 {
        *paths += 1;
    }
    recursive_callers(...);
}
```

PR #22, opened 2018-02-20
fixing computing indices wrong

### LINKED ISSUE #16: Computing some function indices wrong...
For example:

```
$ svelte paths tests/fixtures/mappings.wasm compute_column_spans
 Shallow Bytes │ Shallow % │ Retaining Paths
───────────────┼───────────┼─────────────────────────────────────────
           299 ┊     0.66% ┊ compute_column_spans
               ┊           ┊   ⬑ func[5]
               ┊           ┊       ⬑ export "by_generated_location"

```

sendilkumarn:
fixes #16 

was able to zero in on the issue. But I am not too sure why other tests are broken because of this. 

Open for discussion.


fitzgen:
@pepyakin, could you take a quick look at this and comment on whether we are computing our table indices properly? Thanks a bunch!

pepyakin:
These looks incorrect: entries of the export section are refering the items by indicies in respective index space. E.g. so export of a function with index N will export function by that N index.
So the former code makes more sense to me.

pepyakin:
As for computing table indicies properly: it seems so!

At least this code looks good for me!

https://github.com/fitzgen/svelte/blob/5e1bcd3404583c05d8427e6900b967a52ec098e9/parser/wasm.rs#L499-L502

Except that

https://github.com/fitzgen/svelte/blob/5e1bcd3404583c05d8427e6900b967a52ec098e9/parser/wasm.rs#L495-L497

I would swap these two, so it would be "table → elem → func", it seems more logical to me because logically table retains the elements (in other words, if I remove the exported table, it should remove all the elements and thus referenced functions, not the other way around).

NikVolf:
There are handy functions in parity-wasm like this one (and couple below):

https://paritytech.github.io/parity-wasm/parity_wasm/elements/struct.Module.html#method.import_count

maybe it can help?

sendilkumarn:
> As for computing table indicies properly: it seems so!

@pepyakin I think this makes sense I will try to hack on this.

>  so it would be "table → elem → func", it seems more logical to me because logically table retains the elements 

Is this an assumption or some restriction?

@NikVolf Thanks, guess that works perfect.

sendilkumarn:
> As for computing table indicies properly: it seems so!

@pepyakin I think this makes sense I will try to hack on this.

>  so it would be "table → elem → func", it seems more logical to me because logically table retains the elements 

Is this an assumption or some restriction?

@NikVolf Thanks, guess that works perfect.

sendilkumarn:
@NikVolf well, I need this value `dynamically` (i.e., how much we have covered so far), OTOH computing with total number of `functions | tables | memory...` is not really useful

fitzgen:
@pepyakin,

> I would swap these two, so it would be "table → elem → func", it seems more logical to me because logically table retains the elements (in other words, if I remove the exported table, it should remove all the elements and thus referenced functions, not the other way around).

This makes sense -- thanks for taking a look!

---------------------------------------

@NikVolf,

> There are handy functions in parity-wasm like this one (and couple below):
> 
> https://paritytech.github.io/parity-wasm/parity_wasm/elements/struct.Module.html#method.import_count
> 
> maybe it can help?

What I think would be *really* awesome would be for `parity-wasm` to provide methods that hide all the indexing for us, like:

```rust
impl ElemSegment {
    pub fn table(&self, module: &Module) -> Option<&Table> { ... }
}

impl Func {
    pub fn name(&self, module: &Module) -> Option<&str> { ... }
}

// etc...
```

sendilkumarn:
@fitzgen that makes sense 👍

fitzgen:
@sendilkumarn, are you going to pick this back up and implement what @pepyakin suggested? It might be easier to start from scratch than to rebase this branch...

sendilkumarn:
@fitzgen I swapped the things. But do you mean just swap and no other changes?

fitzgen:
> But do you mean just swap and no other changes?

Yes. My understanding based on the above conversation is that the other changes to the indices were not correct.

Note that the parser underwent a bit of wide reaching changes in https://github.com/fitzgen/svelte/pull/23 so rebasing might be harder than starting from scratch (sorry!)

sendilkumarn:
No issues at all. Let me close this and open another PR with those changes 👍

PR #23, opened 2018-02-21
Use index based Id (and a few other details)

s3bk:
- use (section index, entry index) as Id instead of pointer
- create a map of data sections
- try linking functions to data sections by looking for loads from a constant address

### DIFF
--- a/ir/ir.rs
+++ b/ir/ir.rs
@@ -13,7 +13,7 @@ mod graph_impl;
 use frozen::Frozen;
 use std::cmp;
 use std::collections::btree_map;
-use std::collections::{BTreeMap, BTreeSet, HashMap};
+use std::collections::{BTreeMap, BTreeSet};
 use std::ops;
 use std::slice;
 use std::u32;
@@ -21,10 +21,10 @@ use std::u32;
 /// Build up a a set of `Items`.
 #[derive(Debug)]
 pub struct ItemsBuilder {
-    id_counter: u32,
     size: u32,
-    parsed: HashMap<*const (), Id>,
+    parsed: BTreeSet<Id>,
     items: BTreeMap<Id, Item>,
+    data:  BTreeMap<u32, (Id, u32)>, // offset -> (id, len)
     edges: BTreeMap<Id, BTreeSet<Id>>,
     roots: BTreeSet<Id>,
 }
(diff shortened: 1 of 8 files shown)


fitzgen:
@sebk, thanks for sending this PR!

I haven't looked at the changes yet, but looking at the failing tests on travis CI, it looks like you forgot to update the test expectations. See https://github.com/fitzgen/svelte/blob/master/CONTRIBUTING.md#testing for some more details about the tests.

ir/ir.rs
```diff
@@ -348,8 +346,24 @@ impl<'a> Iterator for Iter<'a> {
...
+        assert!(section < u32::MAX as usize);
+        Id(section as u32, u32::MAX)
+    }
+    /// create `Id` for a given entry if a given section
+    pub fn entry(section: usize, index: usize) -> Id {
+        assert!(section < u32::MAX as usize);
+        assert!(index < u32::MAX as usize);
+        Id(section as u32, index as u32)
+    }
+    /// the root index
```
fitzgen: Nitpick: please capitalize and punctuate doc comments' sentences, here and elsewhere. Also, newlines between methods.

Also, should note that this is the *meta* root's id, since there can be more than one non-meta root.

ir/ir.rs
```diff
@@ -362,15 +376,15 @@ pub struct Item {
...
+    pub fn new<S, K>(id: Id, name: S, size: u32, kind: K) -> Item
     where
         S: Into<String>,
         K: Into<ItemKind>,
     {
         let name = name.into();
         let demangled = demangle(&name);
         Item {
-            id: Id(u32::MAX),
+            id: id,
```
fitzgen: Nitpick: this should use shorthand

```rust
Item {
    id,
    ...
}
```

ir/ir.rs
```diff
@@ -33,68 +33,66 @@ impl ItemsBuilder {
...
         self.roots.insert(id);
         id
     }
 
     /// Add an edge between the given keys that have already been parsed into
     /// items.
-    pub fn add_edge<K, J>(&mut self, from: &K, to: &J) {
-        let from_id = self.id_for_key(from);
-        let to_id = self.id_for_key(to);
+    pub fn add_edge(&mut self, from: Id, to: Id) {
```
fitzgen: This method should `debug_assert!` that we have items for the `from` and `to` ids. The `id_for_key` method used to do that, but its removal means that we could accidentally create dangling edges and not realize it until way down the line. Better to catch the bug early.

parser/wasm.rs
```diff
@@ -579,17 +654,52 @@ impl<'a> Parse<'a> for elements::CodeSection {
...
+            I32Const(c) => return Some(c), // this was easy
+            Loop(_) | If(_) => {}, // keep going
+            I32Add => return pop_stack(code, pos).and_then(|b| pop_stack(code, pos).map(|a| a + b)),
+            I32Sub => return pop_stack(code, pos).and_then(|b| pop_stack(code, pos).map(|a| a - b)),
+            I32Mul => return pop_stack(code, pos).and_then(|b| pop_stack(code, pos).map(|a| a * b)),
+            _ => return None
+        }
+    }
+    None
+}
```
fitzgen: Nitpick: newline between `impl` blocks.

parser/wasm.rs
```diff
@@ -579,17 +654,52 @@ impl<'a> Parse<'a> for elements::CodeSection {
     }
 }
 
+// pos points behind the last position
+fn pop_stack(code: &[elements::Opcode], pos: &mut usize) -> Option<i32> {
```
fitzgen: Woo! \o/ abstract interpretation!

I'd feel much more confident about the correctness of this method if we had a few simple `#[test]`s that covered each of the kinds of opcodes it can handle, a larger combination of opcodes that it can handle, and finally a test with opcodes it cannot handle.

Did you find that interpreting these opcodes was good enough for all intents and purposes to find global data accesses, or do more complicated sequences actually appear in practice that we don't yet handle with this function?

parser/wasm.rs
```diff
@@ -579,17 +654,52 @@ impl<'a> Parse<'a> for elements::CodeSection {
...
+                I32Const(o) => Some(o as i64),
+                I64Const(o) => Some(o),
+                _ => None
+            });
+
+            items.add_item(ir::Item::new(id, name, size, ir::Data::new(ty)));
+
+            let length = d.value().len();
+            match offset {
+                Some(off) => items.link_data(off, length, id),
```
fitzgen: Nitpick: this should be an `if let Some(off) = ...` rather than a `match`.

ir/ir.rs
```diff
@@ -33,68 +33,66 @@ impl ItemsBuilder {
...
+
+    /// locate the data section defining memory at the given offset
+    pub fn get_data(&self, offset: u32) -> Option<Id> {
+        self.data.range(offset..).next().and_then(|(start, &(id, len))| {
+            if offset < start + len {
+                Some(id)
+            } else {
+                None
+            }
+        })
```
fitzgen: Nitpick: don't include trailing white space. Did you run `cargo fmt --all`? I don't think there would be trailing whitespace if so.

https://github.com/fitzgen/svelte/blob/master/CONTRIBUTING.md#automatic-code-formatting

ir/ir.rs
```diff
@@ -33,68 +33,66 @@ impl ItemsBuilder {
...
+            }
+        })                                         
     }
 
     /// Finish building the IR graph and return the resulting `Items`.
     pub fn finish(mut self) -> Items {
-        let meta_root_id = self.next_id();
-        let mut meta_root = Item::new("<meta root>", 0, Misc::new());
-        meta_root.id = meta_root_id;
+        let meta_root_id = Id(u32::MAX, u32::MAX);
```
fitzgen: Shouldn't this use `Id::root()`?

parser/wasm.rs
s3bk: Sadly the abstract interpretation doesn't actually work in practice, because locals are involved, which are much harder to determine.

ir/ir.rs
s3bk: definity. good catch.

fitzgen:
This fixes the issues:

```diff
diff --git a/parser/wasm.rs b/parser/wasm.rs
index 36eeb18..68023e2 100644
--- a/parser/wasm.rs
+++ b/parser/wasm.rs
@@ -540,18 +540,19 @@ impl<'a> Parse<'a> for elements::ElementSection {
 
         for (i, elem) in self.entries().iter().enumerate() {
             let elem_id = Id::entry(idx, i);
             if let Some(table_idx) = table_section {
                 let entry_id = Id::entry(table_idx, elem.index() as usize);
                 items.add_edge(elem_id, entry_id);
             }
             if let Some(func_idx) = func_section {
+                let num_imported_funcs = module.import_count(elements::ImportCountType::Function);
                 for &f_i in elem.members() {
-                    let f_id = Id::entry(func_idx, f_i as usize);
+                    let f_id = Id::entry(func_idx, f_i as usize - num_imported_funcs);
                     items.add_edge(elem_id, f_id);
                 }
             }
         }
         Ok(())
     }
 }
```

I'll create a new PR soon with my local branch of this PR with this patch applied.

s3bk:
I hope this works now …

fitzgen:
Thanks @s3bk!

PR #36, opened 2018-04-17
Add a useful WebAssembly API

fitzgen:
Returns the results of the analyses as a JSON string.


opt/definitions.rs
```diff
@@ -0,0 +1,114 @@
+// Fun times ahead!
+//
+// Apparently, proc-macros don't play well with `cfg_attr` yet, and their
```
alexcrichton: FWIW the [upstream bug](https://github.com/rust-lang/rust/issues/48644) should be closed now so once we have a new nightly this should be working again

alexcrichton:
That's the theory at least!

PR #39, opened 2018-04-18
parser: wasm: fix indexing from exports section to functions/globals/etc sections

fitzgen:
@pepyakin, does this look right? It seems to produce the correct results, AFAICT, but having a second pair of eyes on it would be appreciated :)


fitzgen:
Thanks!

PR #50, opened 2018-05-09
Add a `twiggy garbage` command. Fixes issue #48.

### LINKED ISSUE #48: Add a `twiggy garbage` subcommand to find unreachable items
This would print code and data that is not transitively referenced by any exports / public functions. Since linkers and compilers tend to remove dead code, this is more likely to find bugs in our IR graph, but that is also a useful thing ;)

data-pup:
This fixes #48, and adds a `garbage` subcommand. This includes a test fixture in the form of a wasm module, the source of which can be found [here](https://raw.githubusercontent.com/data-pup/wat-2-wasm-exercises/master/twiggy-garbage-test-fixtures/garbage_test.wat).

Let me know if there are any changes or extra options we should add, I'd be happy to do so. Thanks for the mentoring instructions on this issue, had a lot of fun getting this done! Twiggy is a really neat project :D


analyze/analyze.rs
```diff
@@ -801,3 +799,84 @@ pub fn diff(
...
+        }
+
+        Ok(())
+    }
+}
+
+/// Find items that are not transitively referenced by any exports or public functions.
+pub fn garbage(
+    _items: &ir::Items,
+    _opts: &opt::Garbage,
```
fitzgen: Variables with a leading underscore are usually reserved for unused variables. Stylistically, it is a little weird for these parameters to be used despite the leading underscore. These should be changed to plain `items` and `opts`.

analyze/analyze.rs
```diff
@@ -801,3 +799,84 @@ pub fn diff(
...
+) -> Result<Box<traits::Emit>, traits::Error> {
+    fn get_reachable_items(_items: &ir::Items) -> BTreeSet<ir::Id> {
```
fitzgen: Ditto.

analyze/analyze.rs
```diff
@@ -801,3 +799,84 @@ pub fn diff(
     let diff = Diff { deltas };
     Ok(Box::new(diff) as Box<traits::Emit>)
 }
+
+#[derive(Debug)]
+struct Garbage {
+    items: Vec<ir::Id>,
+    opts: opt::Garbage,
```
fitzgen: Is the `opts` field ever used? If not, then we should remove it.

twiggy/tests/expectations/garbage
```diff
@@ -0,0 +1,11 @@
+ Bytes │ Size % │ Item
```
fitzgen: Can you check in the WAT source used to generate `garbage.wasm` too? That way we can easily make changes if we need to update it in the future.

twiggy/tests/expectations/garbage
```diff
@@ -0,0 +1,11 @@
...
+───────┼────────┼────────
+    11 ┊ 13.10% ┊ code[2]
+     8 ┊  9.52% ┊ code[1]
```
fitzgen: We should probably also build `garbage.wasm` with the `--debug-names` flag, so we get function names instead of `code[N]`.

README.md
```diff
@@ -364,6 +365,15 @@ and new versions of a binary.
         +145 ┊ <wee_alloc::neighbors::Neighbors<'a, T>>::remove::hc9e5d4284e8233b8
 ```
 
+#### `twiggy garbage`
+
+The `twiggy garbage` sub-command finds and display code and data that is not
+transitively referenced by any exports or public functions.
+
+```
+AWOO : Place output example here.
```
fitzgen: Need to fill this out ;)

analyze/Cargo.toml
```diff
@@ -15,3 +15,4 @@ path = "./analyze.rs"
 twiggy-ir = { version = "0.1.0", path = "../ir" }
 twiggy-opt = { version = "0.1.0", path = "../opt", default-features = false }
 twiggy-traits = { version = "0.1.0", path = "../traits" }
+petgraph = "0.4.12"
```
fitzgen: Right now we depend on an older point release in `ir/Cargo.toml`:

```
ir/Cargo.toml:17:petgraph = "0.4.11"
```

Can you update that so that we have the same dependency? Thanks!

analyze/analyze.rs
```diff
@@ -801,3 +799,84 @@ pub fn diff(
...
+}
+
+impl traits::Emit for Garbage {
+    fn emit_text(&self, items: &ir::Items, dest: &mut io::Write) -> Result<(), traits::Error> {
+        let mut table = Table::with_header(vec![
+            (Align::Right, "Bytes".to_string()),
+            (Align::Right, "Size %".to_string()),
+            (Align::Right, "Item".to_string()),
```
fitzgen: This column should be titled "Garbage Item" just to help distinguish the output of this command.

analyze/analyze.rs
data-pup: Oh shoot! I meant to change those, I had placed the garbage function declaration in with an `unimplemented` macro when I was working on the plumbing elsewhere. Will do!

data-pup:
Thanks for the review! I can take care of these changes tonight :)

data-pup:
I added the patches from your review @fitzgen  :) I left this as two separate commits so you can review it faster, but I can rebase these into a single commit if you would like!

Issue #15, opened 2018-02-16
Add ability to display specific subtree of dominators tree


fitzgen:
`$ svelte dominators $FUNCTION` should display just the given function's subtree in the dominator tree.

fitzgen:
@data-pup go for it!

data-pup:
This was solved in #51. Closing this now :)

fitzgen:
FYI, if you put "fixes #1234" in a commit message or PR description, it will automatically close the issue when merged :)

data-pup:
Ahh thanks! I had been tagging them, but it's good to know that there is a specific syntax for that :+1:

PR #51, opened 2018-05-10
Printing dominator subtrees for functions

data-pup:
This commit aims to fix issue #15. Subtrees of the dominators tree can now be printed, if the `--function` flag is given to the `dominator` sub-command. This also works when emitting JSON! This commit also adds a test case for JSON output of the overall dominators tree.

It may be worth noting that if the function name is not found in the dominators tree, then the meta root is used, following the default behavior. I also was not quite sure what to call this flag, we can totally change the name of it if you would like :) Let me know if there are any issues you spot!


data-pup:
Ahh. I think I know how to fix this. Shouldn't take too long, I don't think.

opt/definitions.rs
```diff
@@ -139,6 +139,10 @@ pub struct Dominators {
     /// The maximum number of rows, regardless of depth in the tree, to display.
     #[structopt(short = "r")]
     max_rows: Option<u32>,
+
+    /// The name of the function whose dominator subtree should be printed.
+    #[structopt(long = "function", default_value = "")]
+    func_name: String,
```
fitzgen: Can this be an optional positional argument, ie `twiggy dominators input.wasm my_function`? Unsure if that will be easy to do in structopt or not, but if we can do it without too much hassle, then I think that would be a little nicer ergonomics for users.

I'd also name this "subtree" rather than "func_name".

analyze/analyze.rs
```diff
@@ -358,8 +363,25 @@ pub fn dominators(
     items.compute_dominator_tree();
     items.compute_retained_sizes();
 
+    /// Get the Id of the item with the given name.
+    fn get_item_id(items: &mut ir::Items, name: &str) -> Option<ir::Id> {
```
fitzgen: This seems like a generally useful utility method to add to `ir::Items`:

```rust
impl Items {
    ...

    fn get_item_by_name(&self) -> Option<&Item> {
        ...
    }
}
```

Then, if we just need the id, we can map over the returned option like `items.get_item_by_name("blah").map(|item| item.id())`.

fitzgen:
Oh! And after this merges, would you mind filing a follow up issue to allow the subtree name to be a regex, similar to #58 ? Thanks!

data-pup:
Sure thing! I can take care of those fixes, shouldn't take too long. Once that's done and this has landed, I'll file that issue, wouldn't mind taking a swing at that too :D

PR #63, opened 2018-05-14
Print all paths if no argument is given to `twiggy paths`

### LINKED ISSUE #57: `twiggy paths` should default to everything if no specific items are given
Maybe a good next issue to tackle for you, @data-pup

data-pup:
This is some initial work to try and solve #57. If no argument is given, then all paths should be printed. I was thinking about this, and thought that it might be nice to have this _descend_ the paths rather than printing a whole list of ascending call paths.

I'm not totally sure if that is a good idea or not, and thought I would open up a PR for some thoughts. I have a little bit of refinement work to do on this, right now `type[n]` items are printed out, which I don't think should be included in the output if I understand this correctly.

So, questions I have at the moment:
*  Does descending the call paths make sense if no argument is given?
*  Should `type` items be included in the output? If not, is there an easy way to identify these?
*  How should these be sorted? Name, depth, size? That might be worth adding a new issue for.
*  If the ascending/descending choice is exposed as an option flag, what should that be called?


data-pup:
Oh, some disclaimers: The `successors` tree thing does not really make sense, and I'll need to take that out. Was trying to see if that was a way around the issue of `type` items showing up. The tests I added don't pass right now, but I wanted to open this up & get some thoughts before I push further in this direction :)

fitzgen:
> I was thinking about this, and thought that it might be nice to have this descend the paths rather than printing a whole list of ascending call paths.

Not sure I follow, but I expect looking at the code will clear this up.

> I have a little bit of refinement work to do on this, right now type[n] items are printed out, which I don't think should be included in the output if I understand this correctly.

The thing is that "call graph" is a slight misnomer, and maybe we should fix it; I'd like to hear your take on it. If we were really working with a *call graph* then the only vertices we would have would be functions. But we want (and right now have) vertices for *anything taking up space in the binary*. This includes data segments and custom sections, as well as each type entry. Instead of having edges only where one function calls another function, we also have edges from X to Y whenever X requires that Y also be present in the binary. For example, if you have a function in the binary, it requires that its type entry also exist, so we will make an edge from the function to the type entry.

I'm not 100% sure what to call this graph. "Retaining graph", because X->Y means that X retains Y in the binary? It also is vaguely reminiscent of GC heap graph tooling, which is where a lot of `twiggy`'s inspiration comes from...

> Does descending the call paths make sense if no argument is given?

Need to read the code before I can properly answer.

> Should type items be included in the output? If not, is there an easy way to identify these?

Yes, I think they should be for now. I'd like to avoid any special casing kinds of items, and instead implement a generic way to summarize items in the graph and condense things.

> How should these be sorted? Name, depth, size? That might be worth adding a new issue for.

Definitely by shallow size. Optionally by retained size if a flag is set.

> If the ascending/descending choice is exposed as an option flag, what should that be called?

Again, need to read the code to understand this better.

data-pup:
```
The thing is that "call graph" is a slight misnomer, and maybe we should fix it; I'd like to hear your take on it. If we were really working with a call graph then the only vertices we would have would be functions. But we want (and right now have) vertices for anything taking up space in the binary. This includes data segments and custom sections, as well as each type entry.
```

Ahh, that's really good to know. I think that helps a lot.

```
Yes, I think [type items] should be [included] for now. I'd like to avoid any special casing kinds of items, and instead implement a generic way to summarize items in the graph and condense things.
```

:pray: Awesome. There's definitely some cruff in this commit that comes from experimenting with ways to keep those out of the output, so that is good to know.

data-pup:
Awesome! So, in the case that someone just runs a plain `twiggy paths $THING` with no function names, should that also default to ascending? We could fill the `Paths.items` member with all of the leaf nodes that are `Code` items in that case. Otherwise, we could descend down to `max_depth` etc., starting from the roots :)

I'll start working on this, I'll merge more changes in and let you know when this is ready for further review.

fitzgen:
> So, in the case that someone just runs a plain `twiggy paths $THING` with no function names, should that also default to ascending?

Yes, default to the current ascending behavior.

> We could fill the Paths.items member with all of the leaf nodes that are Code items in that case.

So do an actual graph traversal such that each item is only ever printed once in the output, you're saying? We could do this, but then we would also have to start emitting the size information for every item along a path. I actually like the current behavior of a list of each item in the graph, and then each item has the tree of callers fanning out below it. I'm OK with duplicates across each tree because the way I use the tool is to look at the biggest functions, and then say "huh, who is calling this function?" And inspect just that call tree before moving to the next item and inspecting its call tree, etc.

Or maybe I am misunderstanding something you were saying?

analyze/analyze.rs
```diff
@@ -518,20 +531,39 @@ impl traits::Emit for Paths {
...
 pub fn paths(items: &mut ir::Items, opts: &opt::Paths) -> Result<Box<traits::Emit>, traits::Error> {
-    items.compute_predecessors();
+    if !opts.descending() {
+        items.compute_predecessors();
+    }
 
-    let mut paths = Paths {
-        items: Vec::with_capacity(opts.functions().len()),
-        opts: opts.clone(),
+    let functions: Vec<ir::Id> = match opts.functions().is_empty() {
```
data-pup: Changed this to a Vec<ir::Id> rather than a BTreeSet so that the items could be sorted by size :)

data-pup:
I think I got this working now :) Pardon the lack of clarity in the previous message. Thanks for explaining the purpose of the command a little more, that helped me get a better understanding of what to do. Let me know if there are any details I should change!

data-pup:
Yay! No problem, it's always fun to have an excuse to write wasm by hand :)

Issue #59, opened 2018-05-11
`twiggy dominators` should work with regexes


data-pup:
Once #51 lands, we should add functionality that allows the subtree argument given to `twiggy dominators` to be a regex.

Question: This might also require implementing the ability to print multiple subtrees, is that correct? cc: @fitzgen

PR #68, opened 2018-05-23
Added regex functionality to the dominators sub-command.

data-pup:
Regex commands can now be given to the `dominators` sub-command :o)

This required changing the `root_id` member of the `DominatorTree` struct into a `Vec<ir::Id>` collection, and some minor changes to the `Emit` trait logic. This generally follows the same patterns as the `paths` command, meaning that regular expressions are enabled manually using the `--regex` option.

 Solves #59.

### DIFF
--- a/opt/definitions.rs
+++ b/opt/definitions.rs
@@ -134,7 +134,7 @@ pub struct Dominators {
     output_format: traits::OutputFormat,
 
     /// The name of the function whose dominator subtree should be printed.
-    subtree: Option<String>,
+    items: Vec<String>,
 
     /// The maximum depth to print the dominators tree.
     #[structopt(short = "d")]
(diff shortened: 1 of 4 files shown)


opt/definitions.rs
```diff
@@ -143,6 +143,20 @@ pub struct Dominators {
...
     max_rows: Option<u32>,
+
+    /// Whether or not `items` should be treated as regular expressions.
+    #[structopt(long = "regex")]
+    using_regexps: bool,
+}
+
+impl Dominators {
+    // TODO: wasm-bindgen does not support sending Option<String> across
+    // the wasm ABI boundary yet.
```
fitzgen: It is unclear to me how this TODO is relevant? Is this copy-pasted?

data-pup: Good eye! This comment was mimicking what I saw in the implementation for `Paths` in that file. The comment should say: `// TODO: wasm-bindgen does not support sending Vec<String> across the wasm ABI boundary yet.`

I refactored the `subtree` member in this file from an `Option<String>` into a `Vec<String>` so that it could support multiple arguments in this commit. This was definitely a copy paste error :grimacing:

I fixed this and squashed it into the commit, so it should be good to go now :)

Issue #56, opened 2018-05-11
Each output format should be gated behind a cargo feature


fitzgen:
The twiggy binary should enable them all by default. The wasm API should enable only the JSON output format.

fitzgen:
@csmoe maybe this is something you might be interested in picking up, since you're now intimately familiar with how different formats are emitted :)

csmoe:
ok, I'll take care of this.

csmoe:
@fitzgen do you mean this: `cargo twiggy --features json|csv|...`

PR #70, opened 2018-06-10
[WIP] cargo twiggy

csmoe:
Address #56

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-/target
+**/target
 **/*.rs.bk
 Cargo.lock
 twiggy/tests/whatever-output.txt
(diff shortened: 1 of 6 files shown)


csmoe:
@fitzgen what's use-case of `cargo twiggy ...`? I need some tips about how to test `cargo-twiggy`.

fitzgen:
I actually don't think it is worth pursuing `cargo twiggy` yet.

The motivation was that it would automatically build the crate for wasm and then find the resulting wasm artifact and run twiggy on it. But just building a crate for wasm involves so many custom post-build steps that aren't standardized that I don't think we can easily make `cargo twiggy` too useful right now.

> Address #56

This is not the `cargo twiggy` issue, but I do think that #56 would be a good issue to tackle next!

Sorry for the mix up; I'll close the `cargo twiggy` issue for now.

PR #74, opened 2018-07-02
Added experimental ELF/Mach-O/PE compatibility using gimli/object.

### LINKED ISSUE #69: Add compatibility with ELF binaries
Tracking issue for adding ELF compatibility to twiggy.

#4 lists some binary formats that we would like to add compatibility for, including ELF, Mach-O, and PE/COFF. These might be somewhat involved tasks, so this issue will specifically track work/discussion related to adding ELF compatibility.

As mentioned in that issue, we can use [goblin](https://crates.io/crates/goblin) to parse IR nodes.

[gimli](https://crates.io/crates/gimli) is a library for consuming the DWARF debugging format, which may come in handy. [This](http://dwarfstd.org/doc/Debugging%20using%20DWARF-2012.pdf) is a good introduction link to DWARF. 

The [object](https://crates.io/crates/object) crate might also be useful for this, it is a "unified interface for parsing object file formats."

data-pup:
This PR aims to introduce the ability to parse subroutines from an ELF/Mach-O binary, using the gimli/object crates.

Fixes #69.


data-pup:
Travis failed on the nightly's wasm job, with this error: 

```
error: Could not compile `miniz-sys`.
warning: build failed, waiting for other jobs to finish...
error: build failed
```

ir/ir.rs
```diff
@@ -513,6 +513,15 @@ pub enum ItemKind {
 
     /// Miscellaneous item. Perhaps metadata. Perhaps something else.
     Misc(Misc),
+
+    /// Program scope. Modules, namespaces, imported declarations.
+    Scope(Scope),
+
+    /// Subroutine item.
+    Subroutine(Subroutine),
```
fitzgen: How is this intended to be different from `Code`? I would have expected parsing a `DW_TAG_subroutine` to create a `Code` IR item kind.

ir/ir.rs
```diff
@@ -513,6 +513,15 @@ pub enum ItemKind {
 
     /// Miscellaneous item. Perhaps metadata. Perhaps something else.
     Misc(Misc),
+
+    /// Program scope. Modules, namespaces, imported declarations.
+    Scope(Scope),
```
fitzgen: Scopes only fuzzily exist within a binary, and I'm not sure I see how they are useful for size profiling. Can you explain some more how this is intended to be used?

ir/ir.rs
```diff
@@ -513,6 +513,15 @@ pub enum ItemKind {
...
+
+    /// Subroutine item.
+    Subroutine(Subroutine),
+
+    /// Type entry.
+    Type(Type),
```
fitzgen: Types (ignoring RTTI) don't really exist at runtime or in the binary. Wasm has its types section, but each type entry is pretty tiny and I'd consider it something that could be `Misc`. What is the motivation for this new item kind?

parser/object_parse/compilation_unit_parse/mod.rs
```diff
@@ -0,0 +1,129 @@
...
+        } = extra;
+
+        // Initialize an entry ID counter.
+        let mut entry_id = 0;
+
+        // Find the abbreviations associated with this compilation unit.
+        // Use the abbreviations to create an entries cursor, and move it to the root.
+        let abbrevs = self
+            .abbreviations(&debug_abbrev)
+            .expect("Could not find abbreviations");
```
fitzgen: Ditto regarding `expect` vs `?`.

parser/object_parse/compilation_unit_parse/mod.rs
```diff
@@ -0,0 +1,129 @@
...
+        let mut die_cursor = self.entries(&abbrevs);
+        assert!(die_cursor.next_dfs().unwrap().is_some());
```
fitzgen: Ditto about this assertion.

parser/object_parse/die_parse/item_kind.rs
```diff
@@ -0,0 +1,133 @@
...
+pub fn item_kind<R>(
+    die: &gimli::DebuggingInformationEntry<R, R::Offset>,
+    _debug_types: &gimli::DebugTypes<R>,
+    _compilation_unit: &gimli::CompilationUnitHeader<R, <R as gimli::Reader>::Offset>,
+) -> FallilbleOption<ir::ItemKind>
+where
+    R: gimli::Reader,
+{
+    let item_kind = match die.tag() {
+        gimli::DW_TAG_null => unimplemented!(),
```
fitzgen: All the `unimplemented!()`s in here should just return `None`. It is OK if we don't support everything yet, but we need to just return errors or partial information, rather than panicking.

parser/object_parse/die_parse/item_kind.rs
```diff
@@ -0,0 +1,133 @@
...
+        gimli::DW_TAG_inheritance => unimplemented!(),
+        // Access declarations. (Section 5.7.4)
+        gimli::DW_TAG_access_declaration => None,
+        // Friend entries. (Section 5.7.5)
+        gimli::DW_TAG_friend => unimplemented!(),
+        // Data member entries. (Section 5.7.6)
+        gimli::DW_TAG_member => None,
+        // Variant entries. (Section 5.7.10)
+        gimli::DW_TAG_variant => None,
+        gimli::DW_TAG_variant_part => None,
```
fitzgen: I don't think we should do an exclusive match here, because we explicitly *don't* care about most of these things. It would be easier to do:

```rust
match die.tag() {
    // Match things we care about, and do what we will do with them.
    gimli::DW_TAG_something_we_care_about => do_stuff(die),

    // And then just skip everything else.
    _ => return None,
}
```

parser/object_parse/die_parse/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+use self::location_attrs::DieLocationAttributes;
+
+/// This type alias is used to represent an option return value for
+/// a procedure that could return an Error.
+type FallilbleOption<T> = Result<Option<T>, traits::Error>;
+
+/// This struct represents the extra items required by the Parse trait's
+/// `parse_items` method. This is constructed by the compilation unit's
+/// own implementation of `parse_items`.
+pub struct DIEItemsExtra<'unit, R>
```
fitzgen: Nitpick: Rust style is to only capitalize the first letter of an acronym inside an identifier: `DieItemsExtra`

parser/object_parse/die_parse/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+        let item = match item_kind(self, debug_types, comp_unit)? {
+            Some(kind @ ir::ItemKind::Subroutine(_)) => {
+                let name = item_name(self, debug_str)?
+                    .unwrap_or(format!("Subroutine[{}][{}]", unit_id, entry_id));
+                let id = ir::Id::entry(unit_id, entry_id);
+                DieLocationAttributes::try_from(self)?
+                    .entity_size(addr_size, dwarf_version, rnglists)?
+                    .map(|size| ir::Item::new(id, name, size as u32, kind))
+            }
+            _ => None,
```
fitzgen: If we early return `Ok(())` here, then `item` doesn't need to be an `Option` and we can avoid the `if let Some` below. This would be a nice little code clean up.

parser/object_parse/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+        let arena = Arena::new();
+        let endian = if self.is_little_endian() {
+            gimli::RunTimeEndian::Little
+        } else {
+            gimli::RunTimeEndian::Big
+        };
+
+        // Load the sections of the file containing debugging information.
+        let debug_abbrev: gimli::DebugAbbrev<_> = load_section(&arena, self, endian);
+        let _debug_aranges: gimli::DebugAranges<_> = load_section(&arena, self, endian);
```
fitzgen: If we aren't using the section, then I don't think it makes sense to load it. We can just drop this line.

parser/object_parse/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+        // Identify the file's endianty and create a typed arena to load sections.
+        let arena = Arena::new();
+        let endian = if self.is_little_endian() {
+            gimli::RunTimeEndian::Little
+        } else {
+            gimli::RunTimeEndian::Big
+        };
+
+        // Load the sections of the file containing debugging information.
+        let debug_abbrev: gimli::DebugAbbrev<_> = load_section(&arena, self, endian);
```
fitzgen: Idle thought: it may eventually make sense to make all the sections be `Extra` parameters for *this* parse implementation, and then that way we wouldn't need to duplicate the loads of each section (which potentially means decompressing sections multiple times).

Definitely not something we need to worry about yet, however!

parser/Cargo.toml
```diff
@@ -12,6 +12,10 @@ version = "0.2.0"
...
 [dependencies]
+fallible-iterator = "0.1.4"
+gimli = "0.16.0"
+object = "0.8.0"
 parity-wasm = "0.28.0"
-twiggy-ir = { version = "=0.2.0", path = "../ir" }
-twiggy-traits = { version = "=0.2.0", path = "../traits" }
+typed-arena = "1.3.0"
+twiggy-ir = { version = "0.2.0", path = "../ir" }
+twiggy-traits = { version = "0.2.0", path = "../traits" }
```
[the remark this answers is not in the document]
data-pup: Oops! This was definitely accidental. I ran into some conflicts while rebasing this work before opening this PR, good eye! I'll fix that :)

data-pup:
Thanks for the feedback! Figured I'd answer some of the questions above in a single comment.

Re: `ir::ItemKind` variants, the `Type` kind is definitely superfluous. That's a holdover from some earlier iterations of this work, I can take that out. Same goes for `Scope`, I can remove that as well. Your point about the `Subroutine` variant is also a good one, I can adjust the code so that this creates a `Code` item.

Thanks for the other points, I think those all make a lot of sense! I'll make some edits to take care of those issues, and push a new commit once that's ready. Thanks for the feeback :smile:

parser/object_parse/mod.rs
data-pup: I tried taking a swing at that, and ended up running into some errors regarding unconstrained type parameters. I'm sure this is doable, it's just a little outside of my previous Rust knowledge. Would happily create an issue and follow up on that once the rest of the things here are taken care of, if you think that makes sense :)

data-pup:
Pushed some commits to address the various notes above. The point about features is still a little new to me however. I can read up on those tomorrow and see if I can take care of that detail. Would that be a feature in the `parser/Cargo.toml` file specifically?

parser/object_parse/die_parse/mod.rs
```diff
@@ -56,22 +56,26 @@ where
...
-                    .map(|size| ir::Item::new(id, name, size as u32, kind))
+                if let Some(size) = DieLocationAttributes::try_from(self)?.entity_size(
+                    addr_size,
+                    dwarf_version,
+                    rnglists,
+                )? {
+                    ir::Item::new(id, name, size as u32, kind)
+                } else {
+                    return Ok(());
+                }
```
data-pup: I added early returns here, I think it helped clean things up a *little*. It's worth noting that the `entity_size` method here returns an Option because (if I understand correctly) not all `DW_AT_subprogram` DIEs are describing a definition. These DIEs don't include the same location attributes, in which case this method returns None.

parser/object_parse/mod.rs
fitzgen: Follow up issue is fine :)

parser/Cargo.toml
```diff
@@ -12,10 +12,14 @@ version = "0.2.0"
...
+object = { version = "0.8.0", optional = true }
 parity-wasm = "0.28.0"
-typed-arena = "1.4.0"
+typed-arena = { version = "1.4.0", optional = true }
 twiggy-ir = { version = "=0.2.0", path = "../ir" }
 twiggy-traits = { version = "=0.2.0", path = "../traits" }
+
+[features]
+default = ["dwarf"]
+dwarf = ["fallible-iterator", "gimli", "object", "typed-arena"]
```
fitzgen: Yep, this looks correct. Not sure why builds are hanging now.

parser/object_parse/compilation_unit_parse/mod.rs
```diff
@@ -1,10 +1,12 @@
+#[cfg(feature = "dwarf")]
```
fitzgen: It could be easier to define a macro like:

```rust
macro_rules! if_dwarf {
    ( $( $x:item )* ) => {
        $(
            #[cfg(feature = "dwarf")]
            $x
        )*
    }
}

if_dwarf! {
    // ... definitions and things for when dwarf is enabled go in here ...
}
```

and then you could use it throughout the rest of the crate rather than adding the `#[cfg(...)]`s manually on every single definition.

parser/Cargo.toml
fitzgen: It seems like maybe this was just a Travis CI infra issue -- I've restarted the job to see.

data-pup:
I think I ended up finding a cleaner way around this. Because nothing in `parser/object_parse` would ever be used unless we've enabled the 'dwarf' feature, we can use the cfg attribute on that `mod` statement specifically. This keeps the various `#[cfg(feature = "dwarf")]` statements localized to `parser.rs`, and eliminates the need to have these scattered about in the object file parsing code.

fitzgen:
> Edit: One minor detail I did notice, is that because this work introduces a dependency on gimli to the traits crate, it'd be nice to only include that dependency if we're using the dwarf feature. Not totally sure how to do that atm, but I can cycle back on that once this does build successfully :)

The way to do this is to make `gimli` and `object` be optional dependencies, and have the `dwarf` feature enable them:

```toml
[features]
dwarf = ["gimli", "object"]
```

traits/Cargo.toml
```diff
@@ -13,6 +13,7 @@ path = "./traits.rs"
 
 [dependencies]
 failure = "0.1.1"
+gimli = "0.16.0"
```
fitzgen: This dependency should be optional, and only enabled when the parser has the `dwarf` feature enabled.

traits/traits.rs
```diff
@@ -57,6 +58,15 @@ enum ErrorInner {
 
     #[fail(display = "Regex error: {}", _0)]
     Regex(#[cause] regex::Error),
+
+    #[fail(display = "Gimli error: {}", _0)]
```
fitzgen: And this variant should be `cfg`ed for when `gimli` is enabled.

traits/traits.rs
```diff
@@ -99,6 +110,14 @@ impl From<regex::Error> for Error {
     }
 }
 
+impl From<gimli::Error> for Error {
```
fitzgen: (This is going to need the `cfg` too -- sorry I forgot to point it out earlier!)

data-pup: No problem! Pardon, I should've seen that too. Fixing that now!

PR #84, opened 2018-07-17
Print summary of garbage

matklad:
In https://github.com/rustwasm/team/issues/214#issuecomment-405413733, I was confused by twiggy output, because I hadn't realized that the output is truncated. Let's make it clear and print a summary as well?


matklad:
CI failure looks like a random Tracis flitch

fitzgen:
> CI failure looks like a random Travis glitch.

Restarted the job

Issue #88, opened 2018-07-18
Add summaries to other sub-commands


data-pup:
#84 added a summary to the output of the 'garbage' sub-command, which looks like this:

```
 Bytes │ Size % │ Garbage Item
───────┼────────┼──────────────────────
    11 ┊  5.58% ┊ unusedAddThreeNumbers
     8 ┊  4.06% ┊ unusedAddOne
   ... ┊    ... ┊ ... and 7 more
    43 ┊ 21.83% ┊ Σ
```

I think this is a *really* fantastic idea. Should we add a similar feature to other sub-commands as well? I think the 'top' and 'diff' would be a good start, although I am less sure about the others.

It could be neat to do something similar for the paths command, like this:

```
               Shallow Bytes │ Shallow %          │ Retaining Paths
───────────────┼───────────┼────────────────────────────
            44 ┊    30.56% ┊ "function names" subsection
             7 ┊     4.86% ┊ export "awoo"
               ┊           ┊   ↳ func[4]
               ┊           ┊       ↳ type[0]
               ┊           ┊       ↳ awoo
               ┊           ┊            ↳ ... and 6 more
 ```

(That's imagining what [this file](https://github.com/rustwasm/twiggy/blob/master/twiggy/tests/expectations/paths_test_default_output_desc_with_depth) might look like)

To Do:
* [x] diff
* [x] garbage
* [x] monos
* [ ] paths
* [x] top

fitzgen:
Sounds great to me.

data-pup:
I was thinking more about the comment linked in #84: https://github.com/rustwasm/team/issues/214#issuecomment-405413733

They were using this command: `twiggy garbage -n 99999 main_bg.wasm | wc -l`

Would it be a good idea to add some sort of `--all` flag that would ensure that all items are shown? I don't think that would be too hard to take care of, but I'll look into it and follow up. If that should be in a different issue I can open up a new one for that as well.

fitzgen:
`--all` sounds good to me

PR #94, opened 2018-07-21
Added summary and total rows to top command.

data-pup:
This PR makes a few changes to the `top` sub-command in order to introduce a summary of remaining items when the output is truncated, and a row containing a total of all items. If the retaining size is being used, then these columns will only show "..." in the emitted table.

Other minor changes include changing the name of the `number` field to `max_items`, for the sake of consistency with other sub-commands. This has also been changed from an `Option<u32>` into a plain `u32` value. In order to prevent a breaking change from introducing truncation where there previously was none, this defaults to `u32::MAX`.

I refrained from doing so in this PR, but it seems like a good idea to decide on a consistent default for the maximum number of items to display. `diff` for example, defaults to 20, while others show 10. Once an `--all` flag is introduced, we can go about doing this, and update the test cases as well.

Other note: For now, I kept the `TableRow` struct internal to the text emission function, but this could eventually be refactored out into an entry struct similar to the `DiffEntry` and `MonosEntry` structs shortly, assuming the csv and json emission eventually use it as well.

(This relates to #88)


data-pup:
Rebased this off the new master after merging in #92, due to previous CI errors :)

fitzgen:
> I refrained from doing so in this PR, but it seems like a good idea to decide on a consistent default for the maximum number of items to display. diff for example, defaults to 20, while others show 10. Once an --all flag is introduced, we can go about doing this, and update the test cases as well.

+1 -- also don't think we should be afraid of breaking changes for what is 99% a CLI tool and isn't stable yet at that.

fitzgen:
> Other note: For now, I kept the TableRow struct internal to the text emission function, but this could eventually be refactored out into an entry struct similar to the DiffEntry and MonosEntry structs shortly, assuming the csv and json emission eventually use it as well.

Yeah, some of the earlier `Emit` implementations do a lot more duplicated work in each `emit_*` method than they should.

PR #98, opened 2018-07-26
Total & remaining rows, and csv output for diff.

data-pup:
Continued work for #88. This PR adds a total and remaining row to the output of the diff subcommand, as well as emitting CSV output!

I ended up moving a lot of the duplicated work in the text and json emission into the `diff` function itself, which helped make these much more concise as well :)


fitzgen:
> I ended up moving a lot of the duplicated work in the text and json emission into the diff function itself, which helped make these much more concise as well :)

\o/ Nice!

analyze/analyze.rs
```diff
@@ -1119,6 +1121,18 @@ impl Ord for DiffEntry {
     }
 }
 
+impl serde::Serialize for DiffEntry {
```
fitzgen: Any reason to implement this by hand rather than to use `#[derive(Serialize)]`?

data-pup: Ah! This was implemented manually because the `delta` value should have its sign shown even if it is positive. The derived implementation matched tests for the negative values, but positive values were missing the `+`.

If there's a way to do this with serde macros though, I'm happy to follow up and fix that :)

PR #103, opened 2018-08-02
Add summaries to monos sub-command output.

data-pup:
Continuing on work for #88, adding summaries/total row summaries to the 'monos' sub-command.


analyze/analyze.rs
```diff
@@ -925,81 +964,85 @@ impl traits::Emit for Monos {
...
-            let mut obj = arr.object()?;
-            obj.field("generic", &entry.generic[..])?;
-
-            obj.field(
-                "approximate_monomorphization_bloat_bytes",
-                entry.approx_potential_savings,
-            )?;
-            let approx_potential_savings_percent =
-                (f64::from(entry.approx_potential_savings)) / (f64::from(items.size())) * 100.0;
+        // Give an entry representing a generic function and it various
```
fitzgen: Nitpick: "Give" -> "Given", "and it various" -> "and its various"

analyze/analyze.rs
```diff
@@ -907,16 +907,55 @@ impl PartialOrd for MonosEntry {
...
+            let MonosEntry {
+                name,
+                insts,
+                size,
+                bloat,
+            } = entry;
+
+            let get_size_percent = |x: u32| f64::from(x) / total_size * 100.0;
+
+            let mut rows = vec![TableRow {
```
fitzgen: I think this function could return `-> impl Iterator<Item = TableRow>` and avoid separate heap allocations for each vec. Not a big deal, though.

data-pup: Ooh neat! So, I *think* I took care of this, but I'm not completely sure? I'll hold off on merging this until you have a chance to check on the revised version, but I'd like to try and do it the way you mentioned :)

analyze/analyze.rs
```diff
@@ -939,18 +939,20 @@ impl traits::Emit for Monos {
...
+
             vec![TableRow {
                 bloat: Some(*bloat),
-                bloat_percent: Some(f64::from(*bloat) / total_size * 100.0),
+                bloat_percent: Some(get_size_percent(*bloat)),
                 size: *size,
-                size_percent: f64::from(*size) / total_size * 100.0,
+                size_percent: get_size_percent(*size),
                 name: name.to_string(),
             }].into_iter()
```
fitzgen: The `vec![..]` will still allocate the heap elements, what we want to do is

```rust
use std::iter;

iter::once(TableRow {
    ...
}).chain(...)
```

instead of using `vec![]` here.

data-pup: Omg! Okay *that's* how to do this! I had been having trouble with plain arrays, I didn't know about `iter::once`. I'll take care of that today, thanks for the tip!

data-pup:
Took care of the `vec!` issue mentioned above, and squashed all of these commits down in preparation to merge :smile:

PR #105, opened 2018-08-04
Divided analyzing logic into submodules.

data-pup:
This breaks the different parts of `analyze.rs` into separate files.

The different analyses are grouped together in `analyze::commands`, to differentiate them from the files that are responsible for formatting output into text tables or JSON, which are found in `analyze::formats`. The different analyses, i.e. `top`, `diff`, and so forth are re-exported from `analyze.rs` using a `pub use` statement.

Aside from organizational work, this commit as few changes as possible to the code itself. The only changes made are:
*  For the sake of differentiating between imports coming from other crates within the project, the `extern crate` statements in `analyze.rs` do not alias the other twiggy crates. These *are* aliased when they are imported within the commands files, so that no other code changes.
*  Imports at the top of files within `analyze::commands` are divided into 3 groups, following advice from [this](https://github.com/rust-lang-nursery/fmt-rfcs/issues/24) issue. Imports from standard libraries come first, followed by imports from external crates, and imports from other crates in the project coming third.
*  The `with_header` and `add_row` methods for the `formats::Table` struct have had the `pub` keyword added, so they can be used in the `commands` submodule. This module is not publicly exported from the `analyze` crate however, so I don't believe this has an effect on their visibility elsewhere.


fitzgen:
> The with_header and add_row methods for the formats::Table struct have had the pub keyword added, so they can be used in the commands submodule. This module is not publicly exported from the analyze crate however, so I don't believe this has an effect on their visibility elsewhere.

Just to be double sure, we can mark these as `pub(crate) fn blah()`

fitzgen:
r+ with that change, feel free to merge when its done!

data-pup:
>  am being totally unreasonable, but I would really prefer if the commands submodule were renamed to analyses :-p

Not unreasonable at all! Funnily enough, I had named that submodule `analyses` at the outset, but wasn't originally sure about it. Glad to know my instinct was right! I'll take care of this, and the `pub(crate)` detail mentioned above, and merge this after CI passes :smile: 

Thanks for the review!

Issue #104, opened 2018-08-03
Fix deprecation warning in usage of failure crate


fitzgen:
The new `iter_causes` method does the `skip(1)` dance for us and now `causes` is deprecated too.

```
warning: use of deprecated item 'failure::Fail::causes': please use the 'iter_causes()' method instead
  --> twiggy/./twiggy.rs:22:20
   |
22 |         for c in e.causes().skip(1) {
   |                    ^^^^^^
   |
   = note: #[warn(deprecated)] on by default
```

userzimmermann:
Hi @fitzgen :) Anybody working on this already? If not then please assign to me! It feels like the perfect start into the Rust WebAssembly world :8

PR #106, opened 2018-08-06
main(): Fix deprecated causes() w/iter_causes()

userzimmermann:
Fixes https://github.com/rustwasm/twiggy/issues/104

Hi :) Since I'm completely new to Rust, it took me a while to figure this out... But that's the best way to learn! :D 

Nevertheless I'm still a bit confused ;) `causes()` is defined in `Fail`. Just like `iter_causes()`. The first can be directy used as method on the `Error` trait (even though being deprecated), the second can't... o.O Why?

I also don't know if my solution really works, since I wasn't be able to raise a simple error somehow for testing :D How to do that?

### DIFF
--- a/twiggy/twiggy.rs
+++ b/twiggy/twiggy.rs
@@ -19,7 +19,7 @@ fn main() {
     let options = opt::Options::from_args();
     if let Err(e) = run(&options) {
         eprintln!("error: {}", e);
-        for c in e.causes().skip(1) {
+        for c in Fail::iter_causes(&e) {
             eprintln!("  caused by: {}", c);
         }
         process::exit(1);


twiggy/twiggy.rs
```diff
@@ -19,7 +19,7 @@ fn main() {
     let options = opt::Options::from_args();
     if let Err(e) = run(&options) {
         eprintln!("error: {}", e);
-        for c in e.causes().skip(1) {
+        for c in Fail::iter_causes(&e).skip(1) {
```
fitzgen: Two things:

1. The `iter_causes` method already skips the first error in the chain, so we can remove the `skip(1)` bit now (yay!)
2. Rather than use unified function calling syntax (UFCS) we should use method calling syntax: `e.iter_causes()`

userzimmermann: Hi @fitzgen :) Regarding **1:** Oops! Forgot to re-push after reading that `iter_causes()` _"...does the skip(1) dance for us..."_ ;)

userzimmermann:
Regarding **2** from https://github.com/rustwasm/twiggy/pull/106#discussion_r207965344

Using method calling style `e.iter_causes()` strangely results in:

```
error[E0599]: no method named `iter_causes` found for type `traits::Error` in the current scope
  --> twiggy\./twiggy.rs:22:20
   |
22 |         for c in e.iter_causes() {
   |                    ^^^^^^^^^^^
```

This is that confusing thing I mentioned in my PR description

fitzgen:
Oh wow this is funky, it is a method on the trait object, not a provided trait method: https://docs.rs/failure/0.1.2/src/failure/lib.rs.html#227-229

I... don't know what the impetus for that was.

It should work if we do:

```rust
let e = &e as &Fail;
for c in e.iter_causes() {
    ...
}
```

If that doesn't do the trick, then we can just merge the current version of the PR!

userzimmermann:
I also just fought my way to `trait Fail` and `impl Fail` :D And definitely need to get more understanding about those concepts...

userzimmermann:
@fitzgen, is there no simpler way of casting than `let e = &e as &Fail;`

userzimmermann:
@fitzgen, your cast suggestion does do the trick! :) And can be even simplified:

```rust
for c in (&e as &Fail).iter_causes() {
    ...
}
```

But IMO, `Fail::iter_causes(&e)` looks somehow nicer...

fitzgen:
> But IMO, Fail::iter_causes(&e) looks somehow nicer ;)

agreed

userzimmermann:
Thanks @fitzgen for merging and the nice discussion! But my poll https://twitter.com/zimmermanncode/status/1026540101202059264 about which calling style to use wasn't over yet ;P Luckily everybody voted for `Fail::iter_causes(&e)` so far :D 

I learned a lot from this PR! And will try to get even more involved in the Rust WebAssembly community...

PR #113, opened 2018-08-09
Added remaining stats to the `garbage` command.

data-pup:
Adding stats calculation to the `garbage` sub-command's "... and X more" row. Also add's the remaining/total rows to the JSON output to match the behavior of other sub-commands.

Note: This doesn't clean things up to remove duplicated logic from the `Emit` trait the same way some of the other PR's have, but that can be done soon :smile:


fitzgen:
> Note: This doesn't clean things up to remove duplicated logic from the Emit trait the same way some of the other PR's have, but that can be done soon smile

:+1: :sparkling_heart:

analyze/analyses/garbage.rs
```diff
@@ -71,6 +81,32 @@ impl traits::Emit for Garbage {
...
         }
 
+        match self
+            .items
+            .iter()
+            .skip(self.limit)
+            .map(|id| &items[*id])
+            .fold((0, 0), |(size, cnt), item| (size + item.size(), cnt + 1))
+        {
+            (size, cnt) if cnt > 0 => {
```
fitzgen: Nitpick: I think this would be a bit more readable as

```rust
let (size, cnt) = self
    .items
    .iter()
    .skip(self.limit)
    .map(|id| &items[*id])
    .fold((0, 0), |(size, cnt), item| (size + item.size(), cnt + 1));
if cnt > 0 {
    ...
}
```

PR #116, opened 2018-08-11
test: Compare expected w/actual outputs line-wise

userzimmermann:
Otherwise currently **36/47** tests **FAIL** on Windows due to different line endings, all w/ messages like:

```console
---- ... stdout ----

`twiggy ...` did not have the expected output!
```


twiggy/tests/tests.rs
```diff
@@ -47,11 +47,13 @@ macro_rules! test {
...
             let expected = String::from_utf8_lossy(&expected);
             let expected = expected.trim();
+            let expected_lines = expected.lines().collect::<Vec<&str>>();
 
             let actual = String::from_utf8_lossy(&output.stdout);
             let actual = actual.trim();
+            let actual_lines = actual.lines().collect::<Vec<&str>>();
 
-            if actual != expected {
+            if actual_lines != expected_lines {
```
fitzgen: I think we should also start using

```rust
diff::slice(&expected_lines, &actual_lines)
```

just below this as well.

userzimmermann: But only for performance reasons, right @fitzgen? Since `diff::lines()` already ignores the different line endings...

userzimmermann: I'm also thinking about if `diff::Result::Both()` hunks should be really printed... Since they make it sometimes hard to find the actual difference in large outputs

userzimmermann: And is there an easy way to color those `diff` outputs btw? :D

userzimmermann:
@fitzgen: I applied the `diff::slice()` suggestion :)

Issue #109, opened 2018-08-06
Add an '--all' flag to show all items.


data-pup:
Currently, we are working on adding logic to display summaries of truncated rows for different sub-commands in #88. This is nice because the number of items is displayed, but it would be nice if we had an `--all` option to simply display all items, instead of something like `--max-items $BIG_NUMBER`.

The `top` sub-command would be a nice place to start, but this should eventually be done for all of the subcommands.

To Do:
* [ ] diff
* [ ] dominators
* [ ] garbage
* [ ] monos
* [ ] paths
* [ ] top

data-pup:
@userzimmermann This might be a nice issue for you, if you're looking for other issues to work on here :smile_cat:

userzimmermann:
Definitely, @data-pup! I already created the branch for the PR... :)

userzimmermann:
Don't worry ;) I'm still on it! But first had to deal w/ some confusion when running tests :D #116

rharriso:
Looks like this has been closed?

fitzgen:
> Looks like this has been closed?

The "To Do" checkboxes in the top comment accurately reflect the current state of play. There are a number of commands that don't have an `--all` flag yet.

Black-Tomcat:
Heya all, I'd be willing to tackle this, but I'm new to this whole open source thing. 

Also, for the `top` command, it already has the maximum value for a `u32`
```rust
pub struct Top {
    // ...
    /// The maximum number of items to display.
    #[structopt(short = "n", default_value = "4294967295")]
    max_items: u32,
    // ...
}
```
Would this be fine to keep or if we're adding the all option, should the default be set to something smaller? perhaps 20 like `Diff`?

PR #118, opened 2018-08-12
Add -a, --all flag to garbage command

userzimmermann:
And here it is: The first `--all` flag according to https://github.com/rustwasm/twiggy/issues/109 :)

I also added some new `twiggy garbage` tests for the new flag, based on `fixtures/wee_alloc.wasm`, since `fixtures/garbage.wasm` unfortunately only produces 9 items, but `-n <max_items>` defaults to `10` :D


opt/definitions.rs
```diff
@@ -540,11 +545,12 @@ impl Garbage {
...
     /// The maximum number of items to display.
     pub fn max_items(&self) -> u32 {
-        self.max_items
+        if self.all_items { u32::MAX } else { self.max_items }
     }
 
     /// Set the maximum number of items to display.
     pub fn set_max_items(&mut self, max: u32) {
         self.max_items = max;
+        self.all_items = false;
```
data-pup: Good catch here :smile_cat:

twiggy/tests/tests.rs
```diff
@@ -429,6 +429,32 @@ test!(
...
+
+test!(
+    garbage_wee_alloc_all_json,
+    "garbage",
+    "./fixtures/wee_alloc.wasm",
+    "-f",
+    "json",
+    "-a"
+);
+
```
data-pup: Nice job adding these tests as well!

data-pup:
LGTM! Nice work @userzimmermann! :smile_cat: I see you've also started out on #120, that's awesome! Unless @fitzgen has any other input, I think this can be merged.

fitzgen:
> Unless @fitzgen has any other input, I think this can be merged.

If it gets your approval stamp, that is all I need to know :)

twiggy/tests/tests.rs
userzimmermann: Testing is the most important ;)

userzimmermann:
Thx for nice feedback @data-pup @fitzgen! And approving, merging, etc :) Just right after you merged, it came to my mind that the `...top_10...` expectations should be better named like `...default_10...` or so :D to better understand what they are actually about... What do you think? Worth some extra PR? ;)

PR #123, opened 2018-08-14
Improve failing test output formatting

userzimmermann:
* Don't `.trim()` expected and actual output, since that removes the leading space from the table header lines in the diff view

* Give the output a few more newlines

* Make the actual problems more `.red()` :) using [`colored::Colorize`](https://github.com/mackwic/colored)

You will see how much better it looks as soon as TravisCI is done with this PR ;) since it will fail because of the missing leading table header space in `tests/expectations/monos_maxes` :D (afterwards I will fix that ofc and re-push)

@fitzgen: I hope the little added complexity is not too complex regarding to our discussion in https://github.com/rustwasm/twiggy/pull/116#discussion_r209778370 ;)


userzimmermann:
And there is the red beauty :D https://travis-ci.org/rustwasm/twiggy/jobs/416036607#L617

twiggy/tests/tests.rs
```diff
@@ -59,17 +61,29 @@ macro_rules! test {
...
                 for diff in diff::slice(&expected_lines, &actual_lines) {
                     match diff {
-                        diff::Result::Left(l) => println!("-{}", l),
+                        diff::Result::Left(l) => {
+                            println!("{}", format!("-{}", l).red())
+                        }
                         diff::Result::Both(l, _) => println!(" {}", l),
-                        diff::Result::Right(r) => println!("+{}", r),
+                        diff::Result::Right(r) => {
+                            println!("{}", format!("+{}", r).red())
```
fitzgen: Might as well match normal colorized diff output and use `.green()` for added lines

userzimmermann: @fitzgen: I also thought about `.green()` for the `+` lines, but then I thought: No! `.green()` looks to good for something so wrong ;P 

And those all-red diffs in failing tests are actually quite common in testing frameworks... like [pytest](https://pytest.org)

fitzgen: Ok!

PR #124, opened 2018-08-14
ci: grab the wasm-bindgen version from Cargo.toml

fitzgen:
This way, when we update wasm-bindgen versions, the script doesn't need to also be updated. This will unbreak dependabot for wasm-bindgen updates.

### DIFF
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -2,39 +2,53 @@
 
 set -eux
 
-cd "$(dirname "$0")/.."
+ROOT="$(dirname "$0")/.."
+cd "$ROOT"
 
-case "$JOB" in
-    "test")
-        cargo test --all --exclude twiggy-wasm-api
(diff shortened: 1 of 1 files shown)


fitzgen:
Eventually, this will be done by wasm-pack and everything will be better...

ci/script.sh
```diff
@@ -2,39 +2,53 @@
...
+    esac
+}
+
+function get_wasm_bindgen_version {
+    grep wasm-bindgen -A 1 Cargo.toml \
+        | grep version \
+        | cut -f2 -d '=' \
+        | tr '"' ' ' \
+        | xargs
+}
```
data-pup: :heart_eyes:  love this!

fitzgen: hahaha I'm not proud of this monstrosity

PR #129, opened 2018-08-20
Added regular expressions and specific arguments to diff.

data-pup:
This commit adds the ability to specify items to check the diffs off, both using regular expressions and exact names. Additionally, fixed a minor issue where a "remaining" row would be shown in the output if there were 0 rows remaining.


data-pup:
There's some minor ways I think that the diff function could be cleaned up in the future, but that's being left alone now to help keep this commit clean :)

PR #138, opened 2018-08-24
Added banners to `tests.rs`.

data-pup:
I am planning on adding some extra test cases to help with some other work. This commit adds some simple banner comments and organizes the test based on the command that they are testing. This helps make navigating the tests easier if the calls here are folded.

### DIFF
--- a/twiggy/tests/tests.rs
+++ b/twiggy/tests/tests.rs
@@ -90,6 +90,9 @@ macro_rules! test {
     }
 }
 
+// Top Tests:
+// ----------------------------------------------------------------------------
+
 test!(
     top_wee_alloc,
     "top",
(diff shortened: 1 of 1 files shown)


twiggy/tests/tests.rs
```diff
@@ -90,6 +90,9 @@ macro_rules! test {
     }
 }
 
+// Top Tests:
+// ----------------------------------------------------------------------------
```
fitzgen: Want to split these out into submodules instead?

data-pup: That was the plan, I was having some trouble doing that. Should the `test` macro be duplicated across each file, or is there a way to import that into each submodule after defining it in one place?

fitzgen: You have to add a directory nesting level a little deeper than you might normally expect, similar to `wasm-pack/tests/**`:

```
~/wasm-pack
$ tree tests/all/
tests/all/
├── build.rs
├── main.rs
├── manifest.rs
├── readme.rs
└── utils
    ├── file.rs
    ├── fixture.rs
    ├── manifest.rs
    └── mod.rs
```

The key is that every `tests/*.rs` is its own test crate, so you have to nest into a directory that has a `main.rs` that declares all the modules and all that. The test macro should be able to be defined at the top of `main.rs` and then used in every submodule, I think.

PR #139, opened 2018-08-24
Divided tests out into separate submodules.

data-pup:
This commit divides the `tests.rs` file out into separate submodules for each subcommand, following discussion in #138.


twiggy/tests/all/diff_tests.rs
```diff
@@ -0,0 +1,82 @@
+extern crate colored;
+extern crate diff;
+
+use std::process::Command;
+
+use colored::Colorize;
+
+use slurp;
```
fitzgen: If we do the imports *inside* the function that the test macro expands into, then every test module won't have to have the same seemingly-unused imports.

PR #146, opened 2018-08-28
Reworked `paths` logic, added tests, and fixed a CSV bug.

data-pup:
This commit does a few different things, and largely finishes up the work started in #88, and #140.

First, `analyze/analyses/paths.rs` has been divided into separate files for the `paths` function itself, the emitting logic, and for the struct that represents an entry in the `Paths` struct. Further, the logic for traversing the `ir::Items` tree that was previously mixed into the emitting logic has been divided out into a separate function, `create_entry`.

To be sure that this is all working correctly, I added some various tests as well, especially focusing on the `max_paths` option. The expected output files were generated using the current master branch, to be sure that everything was behaving the same way before and after this change.

While doing this, I *did* notice an issue with the `emit_csv` function, being that the meta root is printed in the `path` field of a CSV record. Additionally, some items only display themselves in that field. In each of these cases, the field will instead be empty. Given that the meta root (iirc) is an internal abstraction, I think it is best left out of this command's output, especially if it is not shown in other formats.

Currently, summaries are not shown, but the code to do so is included in comments under a 'FIXME' note. Once this lands, following up by enabling these sections and updating the test cases should be very straightforward.


analyze/analyses/paths/mod.rs
```diff
@@ -0,0 +1,145 @@
...
+mod paths_emit;
+mod paths_entry;
+
+use self::paths_entry::PathsEntry;
+
+#[derive(Debug)]
+struct Paths {
+    items: Vec<ir::Id>,
+    opts: opt::Paths,
+    entries: Vec<PathsEntry>,
```
data-pup: Whoops, I need to clean up the previous `items` member here, I'll take care of that.

data-pup:
As mentioned in the comment above, I have some light cleanup work to do on this. This should be ready for review very shortly :)

analyze/analyses/paths/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+        opts,
+        entries,
+    };
+
+    Ok(Box::new(paths) as Box<traits::Emit>)
+}
+
+/// This helper function is used to collect ir::Id values for the `items` member
+/// of the `Paths` object, based on the given options.
+pub fn get_items(items: &ir::Items, opts: &opt::Paths) -> Result<Vec<ir::Id>, traits::Error> {
```
data-pup: Nit: This function should be renamed `get_starting_positions`, and have its documentation comment updated.

analyze/analyses/paths/paths_entry.rs
```diff
@@ -0,0 +1,26 @@
...
+pub(super) struct PathsEntry {
+    pub name: String,
+    pub size: u32,
+    pub children: Vec<PathsEntry>,
+}
+
+impl PathsEntry {
+    pub fn _count(&self) -> u32 {
+        1 + self.children.iter().map(|c| c._count()).sum::<u32>()
+    }
```
data-pup: This is included for the sake of the summary logic that can be enabled once this lands :)

twiggy/tests/all/expectations/paths_error_test_no_max_paths_csv
```diff
@@ -0,0 +1,9 @@
+Name,ShallowSize,ShallowSizePercent,Path
+std::io::error::Error::new::h8c006d5367bc92ed,339,0.7494694021931375,func[15] -> std::io::error::Error::new::h8c006d5367bc92ed
+func[15],1,0.0022108241952599928,std::io::impls::<impl std::io::Write for &'a mut W>::write::h5d7e5ba58acd05fd -> std::panicking::LOCAL_STDERR::__getit::h7827294b3348067a -> func[15]
+std::io::impls::<impl std::io::Write for &'a mut W>::write::h5d7e5ba58acd05fd,311,0.6875663247258578,func[33] -> std::io::impls::<impl std::io::Write for &'a mut W>::write::h5d7e5ba58acd05fd
+func[33],1,0.0022108241952599928,elem[0] -> func[33]
+elem[0],59,0.13043862752033958,
+std::panicking::LOCAL_STDERR::__getit::h7827294b3348067a,16,0.035373187124159884,func[37] -> std::panicking::LOCAL_STDERR::__getit::h7827294b3348067a
+func[37],1,0.0022108241952599928,elem[0] -> func[37]
+elem[0],59,0.13043862752033958,
```
data-pup: This expectation file is a good example of duplicate rows in the CSV output. We can clean that up separately :)

fitzgen:
> Minor sidenote: The output generated by emit_csv would probably be most helpful if we made a change so that duplicate rows are not displayed. For now, I've considered that out of scope for this PR, and can open up a separate issue for that.

SGTM

fitzgen:
> Update: This is ready for review now smile_cat

Isn't it totally unfair that there isn't a set of emojis of dog faces with different expressions?!?

data-pup:
> Isn't it totally unfair that there isn't a set of emojis of dog faces with different expressions?!?

You have no idea how often I think about this

analyze/analyses/paths/paths_emit.rs
```diff
@@ -0,0 +1,282 @@
...
+                    |mut res, s| {
+                        res.push_str(s);
+                        res
+                    },
+                )
+        }
+
+        /// Process a given path entry, and return an iterator of table rows,
+        /// representing its related call paths, according to the given options.
+        fn process_entry<'a>(
```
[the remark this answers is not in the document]
data-pup: That makes sense to me! Especially after #105, I agree with you that the nested functions can be adjusted. I'll patch that now.

analyze/analyses/paths/paths_emit.rs
```diff
@@ -280,3 +92,209 @@ impl traits::Emit for Paths {
...
     }
 }
+
+/// This module contains helper functions and structs used by the `emit_text`
+/// method in Path's implementation of the `traits::Emit` trait.
+mod emit_text_helpers {
+    use std::iter;
+
+    use twiggy_ir::Items;
+    use twiggy_opt::Paths;
```
data-pup: Because `Items` and `Paths` are the only things used within these helper modules, and because they are compact, I elected to import these directly rather than using the `twiggy_ir as ir` pattern that is used elsewhere. That pattern helps with readability across a larger file that might use multiple items from another module, but here I think it's fine to be explicit about what we're using :)

analyze/analyses/paths/paths_emit.rs
```diff
@@ -280,3 +92,209 @@ impl traits::Emit for Paths {
...
+
+    use analyses::paths::paths_entry::PathsEntry;
+
+    /// This structure represents a row in the emitted text table. Size, and size
+    /// percentage are only shown for the top-most rows.
+    pub(super) struct TableRow {
+        pub size: Option<u32>,
+        pub size_percent: Option<f64>,
+        pub name: String,
+    }
```
data-pup: I *believe* `pub(super)` is what I should be using here. If that looks wrong though, lmk!

analyze/analyses/paths/mod.rs
```diff
@@ -128,6 +130,8 @@ fn create_entry(
             .collect()
     };
 
+    // Temporarily add the current item to the set of discovered nodes, and
+    // create an entry for each child. Collect these into a `children` vector.
```
data-pup: I added some comments like this in the `create_entry` function, just to clarify what is happening here. Because this is where all the recursion/traversal is happening, figured it was worth highlighting with a comment. :)

analyze/analyses/paths/paths_emit.rs
```diff
@@ -1,113 +1,22 @@
...
 
 use formats::json;
 use formats::table::{Align, Table};
 use twiggy_ir as ir;
-use twiggy_opt as opt;
 use twiggy_traits as traits;
 
-use super::paths_entry::PathsEntry;
-use super::Paths;
+use analyses::paths::Paths;
```
data-pup: Decided to be explicit about the path for this, so that it looks similar to the `use` statements within the helper module below. Totally subjective, happy to change this back if the `super::` syntax is preferable.

data-pup:
@fitzgen, I made some edits to fix the notes in your review. Most of the changes were straightforward, but the point about nested functions introduced some changes that I'd want to run by you before merging.

The gist is, I think it would be nice to keep all of the helper functions within the same file, rather than breaking them out into even ~more~ submodules. Conversely, it would be nice if the `paths_emit` file was not a `impl traits::Emit for ..` block followed by a host of various helper functions, given that there are currently 3 formats to consider.

To fix this, I wrapped the helper functions for the `emit_text`, `emit_json`, and `emit_csv` methods in individual `mod { }` blocks. Each exposes a `process_entry` function, and a helper struct if required.

I think this helps keep the `emit_*` methods readable, and clearly delineates each format's helper functions. I figured this is probably worth getting consensus on first however :smile_cat:

data-pup:
Regarding the CI failure, I think this is something upstream? I saw someone mention in the `#rust-wasm` IRC channel that they were seeing the same error occur while running wasm-pack, so we might want to restart the build once that is fixed :)

PR #153, opened 2018-09-06
Root table items; reverse table/elem edge directions

fitzgen:
Muuuuch less reported in `twiggy garbage` now :)

And tables are showing up as expected in the dominator trees now!


parser/wasm_parse/mod.rs
```diff
@@ -559,7 +559,7 @@ impl<'a> Parse<'a> for elements::ElementSection {
             let elem_id = Id::entry(idx, i);
             if let Some(table_section_idx) = table_section_idx {
                 let entry_id = Id::entry(table_section_idx, elem.index() as usize);
-                items.add_edge(elem_id, entry_id);
+                items.add_edge(entry_id, elem_id);
```
data-pup: > logically it is the table that owns its
elements. It gives us much more useful data this way: we can see when the table
is heavy in the dominators tree, because a bunch of dynamic indirect calls that
the linker couldn't remove got entrenched in the binary.

It took a second for me to understand this, but :exclamation: omg! :exclamation: 

Would not have thought of this, ! This is so cool :)

PR #167, opened 2018-10-02
chore(deps): Update wasm-bindgen to 0.2.23

fitzgen:


opt/definitions.rs
```diff
@@ -43,9 +43,9 @@ pub enum Options {
 }
 
 /// List the top code size offenders in a binary.
+#[wasm_bindgen]
 #[derive(Clone, Debug)]
 #[derive(StructOpt)]
```
data-pup: I like the reordering here :+1:

fitzgen:
I think the CI fail is just an intermtitent network error.

The reordering apparently became necessary at some point in time, no idea why /me shrugs

fitzgen:
Oh, and thanks for review! :)

PR #182, opened 2018-10-10
Render stringified wasm types in type section

### LINKED ISSUE #177: The item name for wasm type entries should have the stringified type
## 💡 Feature Description

We should render the human readable type along with which entry in the types section it is.

The change needs to happen in the `parse_items` trait method impl for `TypeSection`: https://github.com/rustwasm/twiggy/blob/master/parser/wasm_parse/mod.rs#L248-L249

#### 💻 Example Usage

Right now we get output like:

```
              7 ┊      0.00% ┊ type[4]
              7 ┊      0.00% ┊ type[11]
              6 ┊      0.00% ┊ type[3]
```

but instead we should have output like:

```
              7 ┊      0.00% ┊ type[4]: (i32,  i32) -> i32
              7 ┊      0.00% ┊ type[11]: (i32) -> nil
              6 ┊      0.00% ┊ type[3]: (f64) -> f64
```

toVersus:
In current implementation, 'nil' is outputted in parameter block when the function has no parameters.

Close #177


toVersus:
Thanks for your review!
I will push a PR following your suggestion so that it would make more readable than current output :blush:

PR #186, opened 2018-10-12
Hide (summarize) potentially false-positive garbage by default

### LINKED ISSUE #183: `twiggy garbage` should hide/summarize "garbage" data by default
## 💡 Feature Description

When twiggy garbage reports false positives, they are often data segments. The reason is that code accessing data segments can do arbitrary computations to conjure up a data segment's linear memory address and we only recognize a very limited pattern of `(load (const $addr))`.<sup>*</sup>

To avoid reporting so many false positives, by default we should hide individual unused data segments, and instead summarize them in a single row. If the user provides the `--show-data-segments` flag (open to suggestions on a better flag name!) then we can show all the potentially-false-positive data segments like our current behavior.

The potentially-false-positive data segments should not be included in the Σ row. It should have its own, distinct summary row.

<sup>*</sup> *While we can certainly improve our ability to recognize uses of addresses, we can't find everything without literally executing the whole program (and even that might not find every edge if it doesn't encounter every possible input state...).*

#### 💻 Example Usage

Default.

```
$ twiggy garbage input.wasm
 Bytes │ Size % │ Garbage Item
───────┼────────┼───────────────────
   404 ┊  1.02% ┊ code[42]
...
   802 ┊  1.56% ┊ ... and 25 more
  2294 ┊  4.48% ┊ Σ [35 Total Rows]
   596 ┊  1.16% ┊ 14 potential false positive data segments
```

With `--show-data-segments`. Now the data segments _should_ be included in the Σ row.

```
$ twiggy garbage input.wasm --show-data-segments
 Bytes │ Size % │ Garbage Item
───────┼────────┼───────────────────
   404 ┊  1.02% ┊ code[42]
   209 ┊  0.41% ┊ data[25]
   122 ┊  0.24% ┊ data[0]
   113 ┊  0.22% ┊ data[16]
   102 ┊  0.20% ┊ data[1]
    77 ┊  0.15% ┊ data[28]
    77 ┊  0.15% ┊ data[34]
    73 ┊  0.14% ┊ data[12]
    63 ┊  0.12% ┊ data[13]
    60 ┊  0.12% ┊ data[15]
   802 ┊  1.56% ┊ ... and 25 more
  2294 ┊  4.48% ┊ Σ [35 Total Rows]

```

#### 🙌 How to fix this

* Add new flag to definition of `Garbage` options struct: https://github.com/rustwasm/twiggy/blob/master/opt/definitions.rs

* Code for `twiggy garbage` analysis is in this module: https://github.com/rustwasm/twiggy/blob/master/analyze/analyses/garbage.rs

Cldfire:
Closes #183.

I added a test case with the `--show-data-segments` flag for the only wasm file that contained any `data` segments. Also, I didn't touch the code that emits JSON as the issue made it sound like only CLI output was supposed to be changed. If JSON output needs to be changed in the same way, let me know.


analyze/analyses/garbage.rs
```diff
@@ -23,9 +24,19 @@ impl traits::Emit for Garbage {
             (Align::Right, "Size %".to_string()),
             (Align::Left, "Garbage Item".to_string()),
         ]);
+        let items_iter = self.items.iter().map(|id| &items[*id]);
+        let mut hidden_count = 0;
+        let mut hidden_size = 0;
+
+        for item in items_iter.clone().take(self.limit) {
+            // Skip this item if it is a data segment and we're summarizing them later
+            if !self.show_data_segments && item.name().starts_with("data") {
```
fitzgen: There is a better way to determine if this item is a data segment or not:

* add a getter for an item's kind (we just haven't needed one yet): `pub fn kind(&self) -> &ItemKind` to `impl Item { .. }` in `ir/ir.rs`

* and then here, do something like:

```rust
if let ItemKind::Data(_) = item.kind() {
    // ...
}
```

Cldfire: Wow, I can't believe I missed that! Thanks for pointing it out :)

Cldfire:
Yep, that makes sense! I will work on this and update this PR shortly.

PR #191, opened 2018-10-18
Bump parity-wasm to 0.35.0

fitzgen:
This bump wasn't automatic because some APIs that we use changed.


fitzgen:
Thanks for the review :)

Issue #226, opened 2019-01-12
wasm files must have `.wasm` extension.


Badel2:
## 🐛 Bug Description

twiggy fails to detect wasm files when they don't have a `.wasm` extension.

twiggy version: twiggy-opt 0.3.0

#### 🌍 Test Case

Any file works, for example [this one (5MB)](https://github.com/Badel2/comphdl/raw/24bf9e4b7281703875e4f80194ce02db43092c54/demo/nightly/Cargo.cbf6baa7.cargo-web-398fc556ba2b08aea6f4fadc3880dfdf).

#### 👟 Steps to Reproduce

```
$ twiggy top test_case
 Shallow Bytes │ Shallow % │ Item
───────────────┼───────────┼──────────────────
             0 ┊     0.00% ┊ Σ [0 Total Rows]
$ mv test_case test_case.wasm
$ twitty top test_case.wasm
<expected output>
```

#### 😲 Actual Behavior

File is not detected as wasm.

#### 🤔 Expected Behavior

The file should be detected as wasm by looking at its contents, not by using the filename extension.

fitzgen:
Thanks for filing a bug @Badel2! We have code to fallback to auto-detection of the input binary's type, but we don't have tests for it, and --therfore unsurprisingly-- it looks like that functionality is broken.

fitzgen:
It looks like we mistakenly always try to do the generic-binary-with-DWARF path if there is no extension and the "dwarf" feature is enabled. Instead, we should try and sniff the kind of thing we are looking at, and if it is wasm, do the wasm path, and if not, do the generic-binary-with-DWARF path.

This could be a good first issue for a motivated contributor!

https://github.com/rustwasm/twiggy/blob/master/parser/parser.rs#L39-L47

fitzgen:
Some more details:

* Wasm files begin with a "magic number" that can help sniffing: https://webassembly.github.io/spec/core/binary/modules.html#binary-magic
* As do ELF (https://en.wikipedia.org/wiki/Executable_and_Linkable_Format#File_header) and Mach-O (https://lowlevelbits.org/parsing-mach-o-files/ is quickest reference I can find).

Since we use `object` for non-wasm binaries, I think we can simply check for the wasm magic number ourselves, and then defer any other case to `object` (and `goblin`, which it uses).

sepiropht:
I want to try !

fitzgen:
[...]

PR #227, opened 2019-01-14
fix for parsing wasm files without extention

sepiropht:
[#226](https://github.com/rustwasm/twiggy/issues/226)


fitzgen:
Oops, forgot to save the review comment -- one sec >.<

PR #232, opened 2019-01-31
Skip the cpp demangler when symbols don't have the C++ prefixes

brson:
cpp_mangler doesn't seem to handle plain unmangled C symbols correctly,
which can result in panics during stringification.

This workaround assumes that symbols not beginning with "_Z",
"__Z", or "\_GLOBAL\_" are not C++, and uses them as is.

This change allows twiggy to analyze the TiKV dwarf binary on linux without crashing.

More explanation in the code comments. I ran into this when cpp_demangle tried to stringify "RC4", an OpenSSL function, thinking it was a CtorDtor type. cpp_demangle might be able to fix this upstream, but I don't know enough about it.

r? @fitzgen

### DIFF
--- a/ir/ir.rs
+++ b/ir/ir.rs
@@ -589,6 +589,29 @@ impl Code {
             return Some(sym.to_string());
         }
 
+        // If the Rust demangle failed, we'll try C or C++.  C++
+        // symbols almost all start with the prefixes "_Z", "__Z", and
+        // ""_GLOBAL_", except for a special case.
+        //
+        // Per cpp_mangle::ast::MangledName::parse:
+        //
+        // > The libiberty tests also specify that a type can be top level,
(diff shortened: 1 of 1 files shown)


brson:
If you merge this I could use a new published release.

brson:
Seems like the CI error is nighly-wasm specific:

```
   Compiling twiggy-wasm-api v0.3.0 (/home/travis/build/rustwasm/twiggy/wasm-api)
error: can't perform LTO when compiling incrementally
error: aborting due to previous error
error: Could not compile `twiggy-wasm-api`.
```

Seems unrelated to this patch, so I'd prefer not dealing with that.

fitzgen:
Continuing this in https://github.com/rustwasm/twiggy/pull/233

Issue #176, opened 2018-10-05
Twiggy dominators row limiting seems to be ignored


alexcrichton:
Using [this wasm file](https://gist.github.com/alexcrichton/1c73f54d63bcab63fe9c6bc279d90c3d#file-pulldown_cmark_wasm_bg-wasm) executing `twiggy dominators -r 1` still seems to print all the rows! The `--help` page though indicates that `-r` is the maximum number of rows to display

data-pup:
I believe that this will be fixed when #244 lands! :smiley_cat:

data-pup:
Fixed :)

PR #244, opened 2019-02-19
Add unreachable summary to dominators command

### LINKED ISSUE #175: Twiggy dominators depth 1 doesn't add up to the whole file size
Using [this wasm file](https://gist.github.com/alexcrichton/1c73f54d63bcab63fe9c6bc279d90c3d#file-pulldown_cmark_wasm_bg-wasm) you get:

```
$ twiggy dominators -d 1 foo.wasm
────────────────┼────────────┼────────────────────────────────────────
          84355 ┊     39.68% ┊ export "format"
          16045 ┊      7.55% ┊ "function names" subsection
          10997 ┊      5.17% ┊ table[0]
           3465 ┊      1.63% ┊ func[18]
           1836 ┊      0.86% ┊ func[86]
           1199 ┊      0.56% ┊ func[26]
           1059 ┊      0.50% ┊ func[104]
            965 ┊      0.45% ┊ func[2]
            710 ┊      0.33% ┊ func[96]
            340 ┊      0.16% ┊ func[149]
            327 ┊      0.15% ┊ func[150]
            261 ┊      0.12% ┊ func[158]
            140 ┊      0.07% ┊ func[52]
            140 ┊      0.07% ┊ func[74]
            138 ┊      0.06% ┊ func[155]
            137 ┊      0.06% ┊ func[50]
            125 ┊      0.06% ┊ func[92]
            117 ┊      0.06% ┊ func[119]
            106 ┊      0.05% ┊ func[14]
             87 ┊      0.04% ┊ func[108]
             75 ┊      0.04% ┊ func[146]
             70 ┊      0.03% ┊ func[201]
             63 ┊      0.03% ┊ export "__wbindgen_malloc"
             56 ┊      0.03% ┊ func[199]
             53 ┊      0.02% ┊ func[93]
             42 ┊      0.02% ┊ export "__wbindgen_global_argument_ptr"
             34 ┊      0.02% ┊ export "__wbindgen_free"
             28 ┊      0.01% ┊ export "__indirect_function_table"
             22 ┊      0.01% ┊ export "__heap_base"
             21 ┊      0.01% ┊ export "__data_end"
             12 ┊      0.01% ┊ func[87]
             11 ┊      0.01% ┊ export "memory"
              8 ┊      0.00% ┊ type[10]
              8 ┊      0.00% ┊ global[0]
              7 ┊      0.00% ┊ type[4]
              7 ┊      0.00% ┊ type[11]
              6 ┊      0.00% ┊ type[3]
              6 ┊      0.00% ┊ type[7]
              5 ┊      0.00% ┊ type[5]
              5 ┊      0.00% ┊ type[8]
              5 ┊      0.00% ┊ type[14]
              4 ┊      0.00% ┊ type[0]
              4 ┊      0.00% ┊ type[1]
              3 ┊      0.00% ┊ type[6]
```

The file itself is 212k bytes, but the numbers on the left don't seem to add up to that. Is some data left out by accident perhaps?

data-pup:
Fixes #175.

Note: some of the values still don't add up to the _entire_ size of the file, but we have a separate issue at #248 to track that.


fitzgen:
Let me know when I should take a look at this :)

data-pup:
sure thing! I was going to wait until #245 lands, since that will change some of the test outputs, and then clean up the commit history so it is a little easier to review. I'll ping you once that's taken care of!

guide/src/contributing/testing.md
```diff
@@ -27,3 +27,6 @@ twiggy/tests
 To automatically update all test expectations, you can run the tests with the
 `TWIGGY_UPDATE_TEST_EXPECTATIONS=1` environment variable set. Make sure that you
 look at the changes before committing them, and that they match your intentions!
+
+TIP: You can use `git add -p` to examine individual hunks when staging changes
```
data-pup: This is an unrelated change, but figured it would be nice to tell people who aren't familiar with this git flag how to stage changes individually when they do update test expectations

analyze/analyses/dominators/emit.rs
```diff
@@ -0,0 +1,297 @@
+use std::collections::BTreeMap;
```
data-pup: I moved the `Emit` implementation out into a separate submodule in the same manner as the paths analysis. That file was becoming a little hefty, and the emit logic was upstaging the analysis itself. That's in a separate commit to make reviewing easier :)

twiggy/tests/all/expectations/dominators_summarizes_unreachable_items
```diff
@@ -0,0 +1,6 @@
+ Retained Bytes │ Retained % │ Dominator Tree
+────────────────┼────────────┼────────────────────────────
+             58 ┊     29.44% ┊ "function names" subsection
+             48 ┊     24.37% ┊ "local names" subsection
+             21 ┊     10.66% ┊ export "add"
+             43 ┊     21.83% ┊ [9 Unreachable Items]
```
data-pup: This is the kind of output we'll see now in text format :slightly_smiling_face: I could add a 'Total' row, but didn't want this PR to become too large.

twiggy/tests/all/expectations/dominators_json_prints_multiple_root_items
```diff
@@ -0,0 +1 @@
+{"items":[{"name":"calledOnce","shallow_size":5,"shallow_size_percent":3.4722222222222223,"retained_size":5,"retained_size_percent":3.4722222222222223},{"name":"calledTwice","shallow_size":5,"shallow_size_percent":3.4722222222222223,"retained_size":5,"retained_size_percent":3.4722222222222223}]}
```
data-pup: noticed a problem when arguments were given to this subcommand, where the fields would all be shown side by side. This was interfering with the summary output, so I fixed that in this PR :)

twiggy/tests/all/expectations/dominators_csv_does_not_summarize_garbage_if_all_items_are_reachable
```diff
@@ -0,0 +1,18 @@
+Id,Name,ShallowSize,ShallowSizePercent,RetainedSize,RetainedSizePercent,ImmediateDominator
```
data-pup: This test exists to show that we won't show a row summarizing unreachable items if there are no unreachable items

analyze/analyses/utils.rs
```diff
@@ -0,0 +1,14 @@
+use std::collections::BTreeSet;
+
+use petgraph::visit::Walker;
+
+use twiggy_ir as ir;
+
+pub fn get_unreachable_items<'a>(items: &'a ir::Items) -> impl Iterator<Item = &'a ir::Item> {
```
data-pup: This logic is the same that was used in the garbage subcommand, but has been outlined into a `utils` module, so that we can call it in the dominators analysis too. I am open to feedback on where this should be placed, or what this submodule should be called, since I am always hesitant about `utils `modules that can quickly become catch-all junk drawers :T

analyze/analyses/dominators/mod.rs
```diff
@@ -0,0 +1,66 @@
...
+use twiggy_traits as traits;
+
+use crate::analyses::utils;
+
+mod emit;
+
+struct DominatorTree {
+    tree: BTreeMap<ir::Id, Vec<ir::Id>>,
+    items: Vec<ir::Id>,
+    unreachable_items: Vec<ir::Id>,
```
data-pup: This field is new, this is where we store a list of unreachable items' Ids.

analyze/analyses/dominators/mod.rs
```diff
@@ -0,0 +1,66 @@
...
+        sorted_items
+    } else {
+        arguments
+            .iter()
+            .filter_map(|name| items.get_item_by_name(name))
+            .map(|item| item.id())
+            .collect()
+    };
+
+    let unreachable_items = if arguments.is_empty() {
```
data-pup: we only collect the unreachable items when we're analyzing everything. If an argument was given, we skip this and place an empty vector in the field

analyze/analyses/dominators/emit.rs
```diff
@@ -0,0 +1,297 @@
...
+                items,
+                &self.tree,
+                start_depth,
+                &mut row,
+                &opts,
+                *id,
+            );
+        }
+
+        if !self.unreachable_items.is_empty() {
```
data-pup: here's where we add the unreachable summary, if needed

analyze/analyses/dominators/emit.rs
```diff
@@ -0,0 +1,297 @@
...
+        let mut obj = json::object(dest)?;
+        {
+            let mut arr = obj.array("items")?;
+            for curr_id in &self.items {
+                let mut item = arr.object()?;
+                recursive_add_children(items, &self.opts, &self.tree, *curr_id, &mut item)?;
+            }
+        }
+
+        if !self.unreachable_items.is_empty() {
```
data-pup: ditto, for json

analyze/analyses/dominators/emit.rs
```diff
@@ -0,0 +1,297 @@
...
+                    recursive_add_children(items, opts, dominator_tree, child, wtr)?;
+                }
+            }
+            Ok(())
+        }
+
+        let mut wtr = csv::Writer::from_writer(dest);
+        recursive_add_children(items, &self.opts, &self.tree, items.meta_root(), &mut wtr)?;
+
+        if !self.unreachable_items.is_empty() {
```
data-pup: ditto, for csv

twiggy/tests/all/expectations/dominators_wee_alloc_json
```diff
@@ -1 +1 @@
...
...cent":0.1419950301739439}]}]},{"name":"export \"goodbye\"","shallow_size":10,"shallow_size_percent":0.3549875754348598,"retained_size":55,"retained_size_percent":1.9524316648917288,"children":[{"name":"func[9]","shallow_size":1,"shallow_size_percent":0.03549875754348598,"retained_size":45,"retained_size_percent":1.5974440894568689,"children":[{"name":"goodbye","shallow_size":44,"shallow_size_percent":1.5619453319133831,"retained_size":44,"retained_size_percent":1.5619453319133831}]}]},{"name":"export \"memory\"","shallow_size":9,"shallow_size_percent":0.3194888178913738,"retained_size":11,"retained_size_percent":0.3904863329783458,"children":[{"name":"memory[0]","shallow_size":2,"shallow_size_percent":0.07099751508697195,"retained_size":2,"retained_size_percent":0.07099751508697195}]},{"name":"data[0]","shallow_size":9,"shallow_size_percent":0.3194888178913738,"retained_size":9,"retained_size_percent":0.3194888178913738},{"name":"type[3]: (i32) -> nil","shallow_size":4,"shallow_size_percent":0.1419950301739439,"retained_size":4,"retained_size_percent":0.1419950301739439}]}],"summary":[{"name":"[3 Unreachable Items]","retained_size":1084,"retained_size_percent":38.480653177138805}]}
```
data-pup: the end of this file is where we can see the unreachable summary in json form. it's within a separate 'summary' object, so that we can add other little heuristics later on :)

data-pup:
@fitzgen this is ready for review now! :)

analyze/analyses/dominators/emit.rs
```diff
@@ -0,0 +1,298 @@
...
+            items: &ir::Items,
+            dominator_tree: &BTreeMap<ir::Id, Vec<ir::Id>>,
+            depth: u32,
+            mut row: &mut u32,
+            opts: &opt::Dominators,
+            id: ir::Id,
+        ) {
+            assert_eq!(id == items.meta_root(), depth == 0);
+
+            if *row == opts.max_rows() {
```
fitzgen: Should this also be `>` instead of `==` like the max-depth check below?

analyze/analyses/dominators/emit.rs
```diff
@@ -0,0 +1,298 @@
...
+
+            if depth > opts.max_depth() {
+                return;
+            }
+
+            if depth > 0 {
+                add_text_item(items, depth, id, table);
+            }
+
+            if let Some(children) = dominator_tree.get(&id) {
```
fitzgen: If we got here, shouldn't we have discovered this `id` from something that would force it to be in the dominator tree? Perhaps we can just assert/unwrap here.

analyze/analyses/dominators/emit.rs
```diff
@@ -0,0 +1,298 @@
...
+            );
+        }
+
+        if !self.unreachable_items.is_empty() {
+            let unreachable_items_cnt = self.unreachable_items.len();
+            let unreachable_items_size = self
+                .unreachable_items
+                .iter()
+                .map(|id| &items[*id])
+                .map(|item| item.size())
```
fitzgen: This could be written slightly more simply as

```rust
.map(|id| items[id].size())
```

analyze/analyses/dominators/emit.rs
```diff
@@ -0,0 +1,298 @@
...
+
+        if !self.unreachable_items.is_empty() {
+            let mut summary_obj = obj.array("summary")?;
+            let mut unreachable_items_obj = summary_obj.object()?;
+            let unreachable_items_cnt = self.unreachable_items.len();
+            let unreachable_items_size = self
+                .unreachable_items
+                .iter()
+                .map(|id| &items[*id])
+                .map(|item| item.size())
```
fitzgen: ditto -- perhaps instead of duplicating the unreachable size and percent computation in `emit_text` and `emit_json`, we could push it up into the `DominatorTree` struct so that we don't have to duplicate the code?

analyze/analyses/dominators/mod.rs
fitzgen: Yeah so here is where I'd suggest we store the size and percent of unreachable items, since that is all we ever display and we wouldn't have to duplicate the logic of computing it in each of the `emit_*` methods

analyze/analyses/utils.rs
fitzgen: Yeah, I'm always hesitant as well :) In this case, I think we could leave this in the `garbage` module, but make it `pub(crate)` so that teh dominators analysis can use it.

analyze/analyses/dominators/emit.rs
data-pup: hmmmm, I gave this one a try, and ran into this error:

```
thread 'main' panicked at '"\"function names\" subsection" item missing in dominator tree', src/libcore/option.rs:1008:5
```

This might be worth taking care of w/in a separate issue, but I'm totally with you that the `if let` here shouldn't be required

fitzgen: I think maybe the dominator tree only includes an entry if the thing has children. I vaguely remember this... Anyways, carry on!

data-pup: That _sounds_ right to me too, looking at this:

https://github.com/rustwasm/twiggy/blob/master/ir/ir.rs#L212

analyze/analyses/dominators/mod.rs
```diff
@@ -6,15 +6,21 @@ use twiggy_ir as ir;
...
 
 struct DominatorTree {
     tree: BTreeMap<ir::Id, Vec<ir::Id>>,
     items: Vec<ir::Id>,
-    unreachable_items: Vec<ir::Id>,
     opts: opt::Dominators,
+    unreachable_items_summary: Option<UnreachableItemsSummary>,
+}
+
+struct UnreachableItemsSummary {
```
data-pup: I wrapped this up into a struct, because the size, size percentage, and number of unreachable items are conditionally shown in the output. This seemed more elegant & less error prone than multiple `Option<_>` members, and we can destructure this with an `if let Some(..)` in the emitting functions

data-pup: && good idea re: moving this computation up, it makes the `emit_*` methods much cleaner :smile_cat:

analyze/analyses/dominators/mod.rs
```diff
@@ -47,20 +53,30 @@ pub fn dominators(
...
         items: dominator_items,
-        unreachable_items,
         opts: opts.clone(),
+        unreachable_items_summary: summarize_unreachable_items(items, opts),
     };
 
     Ok(Box::new(tree) as Box<traits::Emit>)
 }
+
+fn summarize_unreachable_items(
```
data-pup: outlined this, now that the unreachable items summary is a struct. seemed nicer that way, given that this isn't the primary function of the `dominators` function

analyze/analyses/garbage.rs
```diff
@@ -169,3 +170,14 @@ pub fn garbage(items: &ir::Items, opts: &opt::Garbage) -> Result<Box<traits::Emi
 
     Ok(Box::new(garbage_items) as Box<traits::Emit>)
 }
+
+pub(crate) fn get_unreachable_items<'a>(
```
data-pup: I moved this back into `garbage`, as mentioned in review. Good call! I think this makes a lot more sense :slightly_smiling_face:

twiggy/tests/all/expectations/dominators_wee_alloc_with_depth_and_row
```diff
@@ -1,46 +1,6 @@
...
-              4 ┊      0.14% ┊ type[3]: (i32) -> nil
-              4 ┊      0.14% ┊ code section headers
-              4 ┊      0.14% ┊ data section headers
-              3 ┊      0.11% ┊ type section headers
-              3 ┊      0.11% ┊ function section headers
-              3 ┊      0.11% ┊ table section headers
-              3 ┊      0.11% ┊ memory section headers
-              3 ┊      0.11% ┊ export section headers
-              3 ┊      0.11% ┊ element section headers
+            378 ┊     13.42% ┊   ⤷ wee_alloc::alloc_with_refill::hb32c1bbce9ebda8e
```
data-pup: #176 seems to have been fixed by the `max_rows` fix mentioned in review :)

analyze/analyses/dominators/emit.rs
```diff
@@ -34,11 +35,7 @@ impl traits::Emit for DominatorTree {
         ) {
             assert_eq!(id == items.meta_root(), depth == 0);
 
-            if *row == opts.max_rows() {
-                return;
-            }
-
-            if depth > opts.max_depth() {
+            if *row > opts.max_rows() || depth > opts.max_depth() {
```
data-pup: minor point, but I combined the two early return conditions, since neither is especially dense and it makes this local function a little vertically smaller

data-pup: & as mentioned, the `>` seems to have fixed a separate issue. maybe sometimes code motion _is_ helpful :P

data-pup:
@fitzgen, I took care of the notes mentioned in review! if it all looks good to you, this is ready to merge :smile_cat:

PR #245, opened 2019-02-20
Switch to `wasmparser` from `parity-wasm`

alexcrichton:
This commit switches the wasm size analysis from the `parity-wasm` crate
to the `wasmparser` crate. The primary downside of `parity-wasm` is that
it doesn't retain offsets into the original file, so the current way
used to see how big a section is on disk is to serialize it and measure
the serialized size. The serialized size, however, is not always equal
to the original size when accounting for differences like leb128
encodings.

The `wasmparser` crate has offset information plumbed throughout
parsing, allowing us to faithfully represent the size of various
constructs losslessly from the original file. The `wasmparser` crate,
unlike `parity-wasm`, doesn't provide an IR to work with. This means
that we have to do a bit more data munging ourselves in a few locations.


alexcrichton:
In working on this, I did change at least one functional bit where a function type previously rendered as `() -> nil` now renders as `() -> ()`, although it also supports multiple return values now!

This also fixes up a few locations and how they handled the various index spaces. I think that imported tables/memories/globals/etc should all be working now in addition to the already working imported functions.

One thing I've found is that this still doesn't account for 100% of a wasm file size. I think that's because of various small markers here and there, for example the leading 8 magic bytes or the "this section is N bytes long" leb128 descriptor (or the section id itself). I was wondering, would it be possible to include these sizes somewhere? For example does twiggy handle something like "this section is 10 bytes long, and the first entry within the section is 8 bytes long" where twiggy knows that the section itself is 2 bytes long since 8 are contained in another item

parser/wasm_parse/mod.rs
```diff
@@ -1,95 +1,121 @@
...
+use wasmparser::{self, Operator, SectionReader, Type};
+
+#[derive(Default)]
+pub struct SectionIndices {
+    type_: Option<usize>,
+    code: Option<usize>,
+    functions: Vec<Id>,
+    tables: Vec<Id>,
+    memories: Vec<Id>,
+    globals: Vec<Id>,
```
fitzgen: Ah ok, so these map from index space to the id we have assigned to the parsed thing? Nice :+1:

parser/wasm_parse/mod.rs
```diff
@@ -99,172 +125,232 @@ impl<'a> Parse<'a> for elements::Module {
...
+        for (i, ty) in iterate_with_size(self).enumerate() {
+            let (ty, size) = ty?;
             let id = Id::entry(idx, i);
-            let size = serialized_size(ty.clone())?;
-            // Get the structure of function signature type.
-            let Type::Function(func_type) = ty.clone();
-            let n_params = func_type.params().len();
-
-            // Third member of expression shows the length of comma separated types of parameter.
-            let mut name = String::with_capacity("type[]: () -> ".len() + 4 + (5 * n_params) + 3);
```
fitzgen: Nooooo not my beautiful useless micro-optimizations!!!!!!

twiggy/tests/all/expectations/diff_wee_alloc
```diff
@@ -13,12 +13,12 @@
          -25 ┊ data[2]
          +15 ┊ hello
          +15 ┊ import env::rust_oom
-         +12 ┊ custom section 'linking'
          -12 ┊ elem[0]
           +8 ┊ global[0]
-          -8 ┊ type[4]: (i32, i32, i32, i32, i32) -> nil
```
fitzgen: TBH, I do prefer `nil` since that is what `wasm-objdump` and `wasm-dis` use.

alexcrichton:
Updated back to `nil`!

data-pup:
this is super neat, nice work :slightly_smiling_face: 

> I think that's because of various small markers here and there, for example the leading 8 magic bytes or the "this section is N bytes long" leb128 descriptor (or the section id itself). I was wondering, would it be possible to include these sizes somewhere?

This is a really good point. I could open up a separate issue about this if you'd like!

alexcrichton:
Sounds good to me yeah, I'll open an issue and fill some things in b/c I don't wanna generate work for you!

https://github.com/rustwasm/twiggy/issues/248

PR #247, opened 2019-02-22
Update gimli

philipc:
The tests still pass, and `twiggy top` still shows something, but otherwise I haven't validated this.

The changes include switching to use `gimli::Dwarf` and `gimli::Unit`. This makes it easier to support DWARF 5.


data-pup:
Hi @philipc! Thank you for opening this :slightly_smiling_face: 

#245 just landed, which makes some substantial changes to the parsing code, but this looks like great work! Sorry to make you handle merging conflicts :grimacing: I'd be happy to help out with that if you run into some trouble.

parser/object_parse/die_parse/item_name.rs
```diff
@@ -12,23 +12,20 @@ use super::FallilbleOption;
...
-        Some(s) => {
-            let name = Some(
-                s
-                    .to_string()? // This `to_string()` creates a `Result<Cow<'_, str>, _>`.
-                    .to_string(), // This `to_string()` creates the String we return.
-            );
-            Ok(name)
-        }
-        None => Ok(None),
+    if let Some(attr) = die.attr_value(gimli::DW_AT_name)? {
```
data-pup: Nice job here, I like the `if let` replacing the `match`

fitzgen:
At this point, @philipc is definitely the person most familiar with `gimli`, and you (@data-pup) are the most familiar with Twiggy's gimli integration, so if this looks good to both of y'all, then I don't feel a burning need to review it myself ;)

parser/object_parse/die_parse/item_name.rs
philipc: I still prefer `match` when the arms are smaller. Even better, rust 1.33 will let us use `map` and `transpose` for this kind of error handling.

philipc:
Rebased.

parser/object_parse/die_parse/item_name.rs
data-pup: Yeah! Really excited for `transpose` to finally land :slightly_smiling_face:

data-pup:
Merged! Thanks for contributing this @philipc :slightly_smiling_face:

fitzgen:
Thank you both :)

PR #266, opened 2019-03-12
Bump fallible iterator to 0.2.0 [Don't Merge]

data-pup:
This manually fixes the problems that caused dependabot's PR #263 to fail. For some reason or another, this seems to also require `gimli-rs` to be using `0.2.0` as well? I was seeing cryptic error messages about `FallibleIterator` not being in scope, until I cloned `gimli` and updated the dependency locally.

The fix itself is adjusting our calls to the `FallibleIterator` trait's adaptor methods, so that the return a `Result<T, _>` rather than a `T`.

This shouldn't be merged until https://github.com/gimli-rs/gimli/pull/407 is merged, so I am expecting CI to fail for now. I'll rebase this / poke it again when that should be good to go :+1:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -154,6 +154,11 @@ version = "0.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "fallible-iterator"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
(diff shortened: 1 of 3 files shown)


fitzgen:
> I was seeing cryptic error messages about `FallibleIterator` not being in scope, until I cloned `gimli` and updated the dependency locally.

Its because the "same" trait from different versions of a crate are distinct traits.

PR #270, opened 2019-03-15
Demangle Function Names

### LINKED ISSUE #174: Dominators with depth 1 seemingly doesn't demangle names
Using `twiggy-opt 0.3.0` on a wasm file as `twiggy dominators -d 1 foo.wasm` (trying to get all the roots of large sizes) prints out a lot of entries as `func [xx]`. I found this somewhat confusing because the wasm binary had the function names subsection, but it turns out that I needed to path `-d 2` because it looks like in the dominator tree there's a function node which dominates the node with the function's name. Using `twiggy dominators -d 2` did indeed print out the function names.

Would it be possible to have the function names come up at depth 1 as well?

data-pup:
Fixes #174.

This is a much larger PR than I expected it to be, and apologies on that regard.  :grimacing: 

I tried a simpler fix here: https://github.com/rustwasm/twiggy/compare/master...data-pup:simple-function-name-demangle-fix

Notably, we need to include the type of item in the output, so that we can differentiate between code and function items. Unfortunately, this meant that the user would need to specify items by their entire name and type/location when using subcommands that accept arguments.

This change, while more involved, allows us to query for items by name, while still showing the demangled names of functions and code blocks! By moving the name of an item from the `ir::Item` struct into the `ir::ItemKind` field, we can handle these properly.

The other notable part is that this involves storing the "annotation" (i.e. `func[1]`) in that struct, and adding an `annotated_name` method which is used when displaying items. This is required because the `diff` command breaks if we use the annotated name, because we can't tell that two items are related if an item moved in the index space.

Because this is a hefty PR, I kept changes in separate commits, to help make reviewing this easier. I'll try and find some time to annotate the important parts over the weekend, or early next week :slightly_smiling_face:


twiggy/tests/all/expectations/dominators_wee_alloc_with_depth_and_row
```diff
@@ -1,6 +1,6 @@
  Retained Bytes │ Retained % │ Dominator Tree
-────────────────┼────────────┼──────────────────────────────────────────────────────
+────────────────┼────────────┼───────────────────────────────────────────────────────────────
             777 ┊     27.58% ┊ "function names" subsection
-            387 ┊     13.74% ┊ func[2]
-            378 ┊     13.42% ┊   ⤷ wee_alloc::alloc_with_refill::hb32c1bbce9ebda8e
+            387 ┊     13.74% ┊ func[2]: wee_alloc::alloc_with_refill::hb32c1bbce9ebda8e
+            378 ┊     13.42% ┊   ⤷ code[2]: wee_alloc::alloc_with_refill::hb32c1bbce9ebda8e
```
data-pup: Here's what we see now when running `dominators` now :smile_cat:

fitzgen:
Sorry for the delay -- I'll get to this tomorrow for sure. It is a bit bigger than I expected it to be :-p

fitzgen:
Sorry for the delay -- I'll get to this tomorrow for sure. It is a bit bigger than I expected it to be :-p

fitzgen:
(Woops, mis-click)

ir/ir.rs
```diff
@@ -468,6 +468,22 @@ impl Item {
...
+                .to_string(),
+            ItemKind::Data(Data { name, .. }) => name.to_string(),
+            ItemKind::Func(func) => func.decorator().to_string(),
+            ItemKind::Debug(DebugInfo { name, .. }) => name.to_string(),
+            ItemKind::Misc(Misc { name, .. }) => name.to_string(),
+        }
+    }
+
+    /// Get this item's decorated name.
+    /// TODO: This needs better documentation.
```
fitzgen: Yes, please :)

twiggy/tests/all/expectations/elf_top_25_hello_world_rs
```diff
@@ -1,29 +1,29 @@
...
-          2158 ┊     0.10% ┊ prof_tdata_destroy_locked
-          2041 ┊     0.10% ┊ stats_print_helper
-          1897 ┊     0.09% ┊ imemalign
+───────────────┼───────────┼─────────────────────────────────────────────────
+          9751 ┊     0.46% ┊ Subroutine[68][471]: stats_arena_print
+          9503 ┊     0.45% ┊ Subroutine[5][16]: output
+          7917 ┊     0.38% ┊ Subroutine[68][131]: je_stats_print
+          7852 ┊     0.37% ┊ Subroutine[52][2800]: mallocx
+          7019 ┊     0.33% ┊ Subroutine[53][4212]: je_arena_boot
+          6221 ┊     0.30% ┊ Subroutine[71][86]: je_malloc_vsnprintf
```
fitzgen: I'm concerned that this is a regression in usability for our output, and not an improvement. It seems that every item has an extra prefix now that is not meaningful or actionable to the user.

ir/ir.rs
```diff
@@ -468,12 +468,25 @@ impl Item {
...
+                code.demangled()
+                    .or_else(|| code.name())
+                    .unwrap_or_else(|| code.decorator())
+                    .to_string()
+            },
+            ItemKind::Data(Data { name, .. }) => name.to_string(),
+            ItemKind::Func(func) => {
+                if let Some(name) = func.name() {
+                    // format!("{}: {}", func.decorator(), name)
+                    func.decorator().to_string()
```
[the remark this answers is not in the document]
data-pup: wooops, good catch. Thanks :sweat_smile:

twiggy/tests/all/expectations/elf_top_25_hello_world_rs
data-pup: Yeah, totally agree. Maybe an alternative would be to represent subroutines in an ELF/etc. file with a different variant that wouldn't display this prefix?

data-pup:
> Even this I am not super happy with, though. I wonder if it makes sense to collapse func[123] and code[123] into the same IR item, since no body is used by more than one func and vice versa. They are essentially the same, even if they are split at teh physical binary level (to allow eager type checking and validation). Twiggy's original goal/design was to try and model exactly what the binary layout was, but in this case, since there is no sharing, we might want to loosen that for usability.

Totally agree w/ you about the ergonomics of this re: output being iffy. I can refactor the output to resemble the format you provided, agree w/ you that the extra info would be best listed after the name (if at all).

I've noticed that the code/func relationship is confusing to people when explaining some of the output of this tool, so maybe collapsing these together _would_ be useful. Happy to wait for Alex's feedback on this. At that point, it might be worth deciding whether I should try solving this w/ a different approach altogether, or refactoring this into something w/ nicer output :slightly_smiling_face: 

Thanks for the review @fitzgen :grey_exclamation:

ir/ir.rs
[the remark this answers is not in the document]
data-pup: Turns out that this _did_ get cleaned up in a later commit, was using this to make sure I was still generating the same output when annotations were removed :+1:

ir/ir.rs
data-pup: Ditto here, the `TODO` was removed, but I still have the same sentiment. Maybe a simple example, something like:

```
/// Get this item's decorated name, i.e. 'foo (code[7])'
```

That might warrant adding some other info to the existing `name` method, which I'm also happy to add if you think that makes sense :slightly_smiling_face:

ir/ir.rs
```diff
@@ -471,12 +468,39 @@ impl Item {
     }
 
     /// Get this item's name.
-    #[inline]
     pub fn name(&self) -> &str {
-        if let ItemKind::Code(ref code) = self.kind {
-            code.demangled().unwrap_or(&self.name)
-        } else {
-            &self.name
+        match &self.kind {
```
data-pup: Another note, while we're in review: I _sort of_ detest that this is called the item's "kind" if it now also stores the name due to this approach. Would this be best referred to as `ItemMetadata` or something along those lines?

twiggy/tests/all/expectations/elf_top_25_hello_world_rs
data-pup: similar to the `Func` variant, other things need not create a `Subroutine` variant if that isn't applicable, and in the case of DWARF info, we can hide that behind a `cfg` flag anyhow

alexcrichton:
I think I agree with @fitzgen that the best way to handle this on WebAssembly is to merge the func/code nodes. When looking at the size of a function we may as well account for the bytes that the function type pointer takes up (although it's like 1-3 at most realistically). It'll definitely make reading the output a bit easier as I know in the past I've had to follow chains a few levels deep to get to the real part of a function!

data-pup:
That sounds good to me! Thanks for the review/advice y'all :tada: :pray: 

I'm going to close this for now, because it is probably easiest to start over if we should go that route. If I run into any trouble I'll ping for advice back in #174.

fitzgen:
Thanks @data-pup! Sorry for the delay followed by the back-and-forth :(

ir/ir.rs
fitzgen: heh, we used `kind` in `rust-bindgen`, so it is familiar to me, but I don't have much of a first-principles argument here

data-pup:
@fitzgen not a problem _at all!_ I know you're all kinda busy w/ gloo & other projects atm :slightly_smiling_face: Most of this work did feel speculative & exploratory anyhow, so it's nice to feel confident about why merging func/code items in the IR model is a better approach than this.

PR #308, opened 2019-06-25
Demangle function names

### LINKED ISSUE #174: Dominators with depth 1 seemingly doesn't demangle names

data-pup:
Fixes #174. (Finally! :smile_cat:)

In #270, we [proposed](https://github.com/rustwasm/twiggy/pull/270#pullrequestreview-216947451) [merging](https://github.com/rustwasm/twiggy/pull/270#issuecomment-475041468) the IR items for code/function section entries.

>  I wonder if it makes sense to collapse func[123] and code[123] into the same IR item, since no body is used by more than one func and vice versa. They are essentially the same, even if they are split at the physical binary level (to allow eager type checking and validation).

> I think I agree [..] that the best way to handle this on WebAssembly is to merge the func/code nodes. When looking at the size of a function we may as well account for the bytes that the function type pointer takes up (although it's like 1-3 at most realistically). It'll definitely make reading the output a bit easier as I know in the past I've had to follow chains a few levels deep to get to the real part of a function!

This combines the func/code nodes in the IR generated when parsing a wasm module :tada: 

To keep this somewhat more reviewable, I've divided the PR into 3 commits.
*  9e82d6c - No changes to behavior, this simply refactors the parsing code towards a structure where merging the func/code nodes can be accomplished in as small a change as possible. This commit will still pass the original test snapshots.
*  474b193 - This is where the actual IR node merge occurs, including changes to how we draw edges between nodes.
*  e1fcfba - This commit is purely updating the test snapshots. This should probably be reviewed _first_, so that we can decide whether this is an acceptable change regarding ergonomics/correctness :slightly_smiling_face:


twiggy/tests/all/expectations/dominators_wee_alloc_with_depth_and_row
```diff
@@ -1,6 +1,6 @@
  Retained Bytes │ Retained % │ Dominator Tree
-────────────────┼────────────┼──────────────────────────────────────────────────────
+────────────────┼────────────┼────────────────────────────────────────────────────
             777 ┊     27.58% ┊ "function names" subsection
-            387 ┊     13.74% ┊ func[2]
-            378 ┊     13.42% ┊   ⤷ wee_alloc::alloc_with_refill::hb32c1bbce9ebda8e
+            387 ┊     13.74% ┊ wee_alloc::alloc_with_refill::hb32c1bbce9ebda8e
+            226 ┊      8.02% ┊   ⤷ wee_alloc::alloc_first_fit::h9a72de3af77ef93f
```
data-pup: Here's the test case that I think demonstrates this change most succinctly. Now, the topmost node contains the name of the function/code. The immediate child is another function that this dominates, as opposed to the code node that was displayed before.

twiggy/tests/all/expectations/garbage
```diff
@@ -1,12 +1,9 @@
  Bytes │ Size % │ Garbage Item
 ───────┼────────┼────────────────────────────────
-    11 ┊  5.58% ┊ unusedAddThreeNumbers
-     8 ┊  4.06% ┊ unusedAddOne
+    12 ┊  6.09% ┊ unusedAddThreeNumbers
+     9 ┊  4.57% ┊ unusedAddOne
```
data-pup: In general, most of the test changes look like this: A single byte being added to the node that now represents both the code _and_ function entries. Other changes, such as the rest of this file, are because there are fewer items listed in the 'total rows' message.

twiggy/tests/all/expectations/paths_test_called_twice
```diff
@@ -1,13 +1,9 @@
...
-               ┊           ┊           ⬑ func[3]
-               ┊           ┊               ⬑ export "woof"
+───────────────┼───────────┼────────────────────────────
+             6 ┊     4.17% ┊ calledTwice
+               ┊           ┊   ⬑ bark
+               ┊           ┊       ⬑ export "bark"
+               ┊           ┊       ⬑ awoo
+               ┊           ┊           ⬑ export "awoo"
+               ┊           ┊   ⬑ woof
+               ┊           ┊       ⬑ export "woof"
```
data-pup: Paths outputs also look much nicer now! Fewer intermediate func[i] nodes.

parser/wasm_parse/mod.rs
```diff
@@ -73,81 +92,77 @@ impl<'a> Parse<'a> for wasmparser::ModuleReader<'a> {
...
                         .get_import_section_reader()?
                         .parse_items(items, idx)?;
-                    "import section headers".to_string()
                 }
                 wasmparser::SectionCode::Function => {
-                    section
-                        .get_function_section_reader()?
-                        .parse_items(items, idx)?;
-                    "function section headers".to_string()
+                    panic!("unexpected function section");
```
data-pup: tangentially, not _entirely_ sure what to do right here. We shouldn't ever hit this branch.

parser/wasm_parse/mod.rs
```diff
@@ -292,6 +325,176 @@ impl<'a> Parse<'a> for wasmparser::ModuleReader<'a> {
...
+        wasmparser::SectionCode::Export => "export section headers".to_string(),
+        wasmparser::SectionCode::Start => "start section headers".to_string(),
+        wasmparser::SectionCode::Element => "element section headers".to_string(),
+        wasmparser::SectionCode::Code => "code section headers".to_string(),
+        wasmparser::SectionCode::Data => "data section headers".to_string(),
+        wasmparser::SectionCode::DataCount => "data count section headers".to_string(),
+    }
+}
+
+impl<'a> Parse<'a> for (IndexedSection<'a>, IndexedSection<'a>) {
```
data-pup: Function/Code sections are merged by parsing them together. This is where most of the important changes to this file are happening.

data-pup:
At the moment, there's still some light cleanup that I'd want to do before this gets merged. However, I wanted to open this up for review to check that the alterations to the output, and general approach to implementing this change, look reasonable :slightly_smiling_face:

twiggy/tests/all/expectations/dominators_regex_any_func
```diff
@@ -1,14 +1,2 @@
...
-              5 ┊      3.47% ┊           ⤷ calledOnce
-              6 ┊      4.17% ┊ func[0]
-              5 ┊      3.47% ┊   ⤷ calledOnce
-              6 ┊      4.17% ┊ func[1]
-              5 ┊      3.47% ┊   ⤷ calledTwice
-              6 ┊      4.17% ┊ func[2]
-              5 ┊      3.47% ┊   ⤷ bark
-              6 ┊      4.17% ┊ func[4]
-              5 ┊      3.47% ┊   ⤷ awoo
+────────────────┼────────────┼───────────────
```
data-pup: this test would need to be updated to use a different regex, since `func[i]` isn't actually a name that shows up anymore!

parser/wasm_parse/mod.rs
```diff
@@ -73,81 +92,77 @@ impl<'a> Parse<'a> for wasmparser::ModuleReader<'a> {
...
         }
 
-        for (idx, (size, section)) in sections.into_iter().enumerate() {
+        // Next, we parse the function and code sections together, so that we
+        // can collapse corresponding entries from the code and function
+        // sections into a single representative IR item.
+        match (function_section, code_section) {
+            (Some(function_section), Some(code_section)) => (function_section, code_section)
+                .parse_items(items, (imported_functions, &names, &sizes))?,
+            _ => panic!("function or code section is missing"),
```
fitzgen: This should be a `failure::bail!("...")` instead of a panic

parser/wasm_parse/mod.rs
fitzgen: If this branch is actually unreachable, as opposed to something that can happen but is an error, then we should use the `unreachable!()` macro. I believe that is the case here, right?

parser/wasm_parse/mod.rs
```diff
@@ -73,81 +92,77 @@ impl<'a> Parse<'a> for wasmparser::ModuleReader<'a> {
...
                         .get_element_section_reader()?
                         .parse_items(items, idx)?;
-                    "element section headers".to_string()
                 }
                 wasmparser::SectionCode::Code => {
-                    section
-                        .get_code_section_reader()?
-                        .parse_items(items, (idx, imported_functions, &names))?;
-                    "code section headers".to_string()
+                    panic!("unexpected code section");
```
fitzgen: Ditto.

parser/wasm_parse/mod.rs
```diff
@@ -274,10 +312,7 @@ impl<'a> Parse<'a> for wasmparser::ModuleReader<'a> {
                         .parse_edges(items, (&indices, idx))?;
                 }
                 wasmparser::SectionCode::Code => {
-                    indices.type_ = Some(idx);
-                    section
-                        .get_code_section_reader()?
-                        .parse_edges(items, (&indices, idx))?;
+                    panic!("unexpected code section found");
```
fitzgen: These should be `unreachable!()` too, I think

parser/wasm_parse/mod.rs
```diff
@@ -73,7 +73,9 @@ impl<'a> Parse<'a> for wasmparser::ModuleReader<'a> {
         match (function_section, code_section) {
             (Some(function_section), Some(code_section)) => (function_section, code_section)
                 .parse_items(items, (imported_functions, &names, &sizes))?,
-            _ => panic!("function or code section is missing"),
+            _ => Err(traits::Error::with_msg(
+                "function or code section is missing",
+            ))?,
```
fitzgen: Ah ok perfect -- I think `failure::bail!` wouldn't work for us anyways since I forgot that we don't use `failure::Error`

parser/wasm_parse/mod.rs
```diff
@@ -122,15 +121,18 @@ impl<'a> Parse<'a> for wasmparser::ModuleReader<'a> {
...
                     section.get_data_section_reader()?.parse_items(items, idx)?;
                 }
                 wasmparser::SectionCode::DataCount => {
                     DataCountSection(section).parse_items(items, idx)?;
                 }
+                wasmparser::SectionCode::Code => {
+                    Err(traits::Error::with_msg("unexpected code section"))?
+                }
+                wasmparser::SectionCode::Function => {
+                    Err(traits::Error::with_msg("unexpected function section"))?
```
fitzgen: These should still be `unreachable!()` however.

data-pup:
Thanks for the review! I resolved your notes, adding `unreachable!` macros to branches that are in fact unreachable, addressed warnings from `clippy`, and rebased this off of current `master`. Once CI finishes up, this should be good to merge :cowboy_hat_face:

fitzgen:
Woo! Thanks @data-pup! I think we should get this in folks' hands -- is there anything else you want to fix before we do the next release?

data-pup:
No problem @fitzgen! Was fun to figure this one out. There isn't anything pressing enough atm to warrant holding off on a release, I think this provides enough improvement to the output that it's worth doing a release sooner than later.

fitzgen:
Published!

* https://users.rust-lang.org/t/announcing-twiggy-version-0-6-0/29708
* https://www.reddit.com/r/rust/comments/c6apb2/announcing_twiggy_version_060/
* https://twitter.com/fitzgen/status/1144344943554519040

PR #365, opened 2019-10-25
fix #326 (help-wanted)

### LINKED ISSUE #326: Panic when running twiggy on multithreaded wasm module
I ran into an assertion when running `twiggy top` on a multithreaded wasm module (attached here)

```
$ twiggy top raytrace_parallel_bg.wasm
thread 'main' panicked at 'should not parse the same key into multiple items', /home/alex/.cargo/registry/src/github.com-1ecc6299db9ec823/twiggy-ir-0.6.0/./ir.rs:53:9
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

[raytrace_parallel_bg.wasm.gz](https://github.com/rustwasm/twiggy/files/3411808/raytrace_parallel_bg.wasm.gz)

sepiropht:
[#326](https://github.com/rustwasm/twiggy/issues/326)
The `add_item` function is called two times with the same id. 
In one iteration we call successively   `DataCountSection(section).parse_items(items, idx)?;` and 
` items.add_root(ir::Item::new(id, name, size - added, ir::Misc::new()));` with the same idx value.
That why we have a panic in the `add_item` function.
I'm suppose that id is the same because section's code is also the same ? I didn't understand further

### DIFF
--- a/ir/ir.rs
+++ b/ir/ir.rs
@@ -50,6 +50,7 @@ impl ItemsBuilder {
         self.items.insert(id, item);
 
         let old_value = self.parsed.insert(id);
+        // panic happen as this is call 2 times with 7,
         assert!(
             old_value,
             "should not parse the same key into multiple items"
--- a/parser/wasm_parse/mod.rs
+++ b/parser/wasm_parse/mod.rs
@@ -125,12 +125,18 @@ impl<'a> Parse<'a> for wasmparser::ModuleReader<'a> {
                     section.get_data_section_reader()?.parse_items(items, idx)?;
                 }
                 wasmparser::SectionCode::DataCount => {
+                    let msg = "the last iteration of the loop take 7 as id";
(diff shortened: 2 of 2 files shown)


fitzgen:
Hey @sepiropht, can you create a test case and steps to reproduce the bug? That would help diagnose the problem and come up with what the correct fix might be. Thanks!

sepiropht:
To reproduce the bug i just used the file that was referenced in the issue [raytrace_parallel_bg.wasm.gz](https://github.com/rustwasm/twiggy/files/3411808/raytrace_parallel_bg.wasm.gz)
`twiggy top raytrace_parallel_bg.wasm`

data-pup:
Hey @sepiropht! @fitzgen added a small test case you can use, over in #326. You should be able to use that to help finish this up. Thanks so much for opening a PR, and sorry about the slow reply!

PR #402, opened 2019-12-25
Use dwarf linkage names

### LINKED ISSUE #380: die_parse::item_name should probably use DW_AT_linkage_name instead of DW_AT_name.
While it's possible to recover the full path by taking into account the full stack of `DW_TAG_namespace`s a `DW_TAG_subprogram` is nested in, it seems easier to rely on `DW_AT_linkage_name` to get a mangled symbol.

For example is an excerpt from `llvm-dwarfdump` on a WASM binary:
```c
0x00003712:     DW_TAG_namespace
                  DW_AT_name    ("fmt")

0x00003722:       DW_TAG_namespace
                    DW_AT_name  ("{{impl}}")

0x00003727:         DW_TAG_subprogram
                      DW_AT_low_pc      (0x0000000000002121)
                      DW_AT_high_pc     (0x0000000000002137)
                      DW_AT_linkage_name        ("_ZN50_$LT$$RF$mut$u20$W$u20$as$u20$core..fmt..Write$GT$9write_str17hbf3f51d4e3d51dd3E")
                      DW_AT_name        ("write_str<alloc::string::String>")
                      DW_AT_decl_file   ("/home/eddy/Projects/rust-1/src/libcore/fmt/mod.rs")
                      DW_AT_decl_line   (203)
```
After replacing `DW_AT_name` with `DW_AT_linkage_name` in `parser/object_parse/die_parse/item_name.rs`, full demangled symbols were available in `--mode=dwarf` output, so it seems to "just work".

data-pup:
Fixes #380.

This makes a small change to `parser/object_parse/die_parse/item_name.rs` so that we first check for a linkage name via `DW_AT_linkage_name`, before looking for an identifier name with `DW_AT_name` (where we currently find name information for a given DIE). We still use identifier names as a fallback, for items that do not have a linkage name.

### DIFF
--- a/parser/object_parse/die_parse/item_name.rs
+++ b/parser/object_parse/die_parse/item_name.rs
@@ -19,7 +19,10 @@ pub fn item_name<R>(
 where
     R: gimli::Reader,
 {
-    die.attr_value(gimli::DW_AT_name)?
+    die.attr_value(gimli::DW_AT_linkage_name)
+        .transpose()
+        .or_else(|| die.attr_value(gimli::DW_AT_name).transpose())
+        .transpose()?
         .map(
             |attr: gimli::read::AttributeValue<R>| -> Result<String, traits::Error> {
                 Ok(
--- a/parser/object_parse/die_parse/item_name.rs
+++ b/parser/object_parse/die_parse/item_name.rs
@@ -3,14 +3,24 @@ use twiggy_traits as traits;
 
 use super::FallilbleOption;
 
-/// Calculate the item's name. For more information about this, refer to Section 2.15 of
-/// the DWARF v5 specification: 'Identifier Names'. Any DIE associated representing an
-/// entity that has been given a name may have a `DW_AT_name` attribute. If there was
-/// not a name assigned to the entity in the source code, the attribute may either not
-/// exist, or be a single null byte.
+/// Calculate the item's name. If no name was assigned, a name will be
+/// decided elsewhere using the `ir::ItemKind` variant that was determined
+/// for the entity.
 ///
-/// If no name was assigned, a name will be decided elsewhere using the
-/// ir::ItemKind variant that was determined for the entity.
+/// For more information on identifier names and linkage names, refer to
+/// Section 2.15 and Section 2.22 of the DWARF v5 specification, respectively.
+///
+/// > _"Any debugging information entry representing a program entity that
+/// > has been given a name may have a DW_AT_name attribute, whose value
+/// > is a string representing the name as it appears in the source program."_
+///
+/// - DWARF v5 Spec. Section 2.15
+///
+/// > _"A debugging information entry may have a DW_AT_linkage_name attribute
+/// > whose value is a null-terminated string describing the object file
+/// > linkage name associated with the corresponding entity."_
+///
+/// -- DWARF v5 Spec. Section 2.22
 pub fn item_name<R>(
     die: &gimli::DebuggingInformationEntry<R, R::Offset>,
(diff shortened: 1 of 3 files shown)


parser/object_parse/die_parse/item_name.rs
```diff
@@ -1,15 +1,26 @@
...
+/// > has been given a name may have a DW_AT_name attribute, whose value
+/// > is a string representing the name as it appears in the source program."_
+///
+/// - DWARF v5 Spec. Section 2.15
+///
+/// > _"A debugging information entry may have a DW_AT_linkage_name attribute
+/// > whose value is a null-terminated string describing the object file
+/// > linkage name associated with the corresponding entity."_
+///
+/// -- DWARF v5 Spec. Section 2.22
```
data-pup: :point_up: Updated the docstring, now that we are using an additional attribute. No real changes to the content itself.

parser/object_parse/die_parse/item_name.rs
```diff
@@ -18,13 +29,19 @@ pub fn item_name<R>(
...
+            |attr: gimli::read::AttributeValue<R>| -> Result<String, traits::Error> {
+                Ok(
+                    dwarf
+                        .attr_string(unit, attr)?
+                        .to_string()? // This `to_string()` creates a `Result<Cow<'_, str>, _>`.
+                        .to_string(), // This `to_string()` creates the String we return.
+                )
+            },
+        )
+        .transpose()
```
data-pup: This is the meat of the PR, and it's admittedly dense. Apologies!

The alternative was outlining the contents of the old `if let` block into a helper function, which was too vertically noisy for my taste, or (_hiss_) duplicating the code into a new `else if let`.

It _is_ a little clever*, but the `transpose` calls let us lazily call `die.attr_value(gimli::DW_AT_name)` only in the case that we have an `Ok(None)`, while still bailing early if we've encountered some bad data. The final `transpose` is to let us use `?`'s in our mapping closure, and bubble the result back up so we're matching the type signature.

* in situations like this, I mean "clever" with the pejorative connotation.

twiggy/tests/all/expectations/elf_top_25_hello_world_rs
```diff
@@ -1,25 +1,25 @@
  Shallow Bytes │ Shallow % │ Item
-───────────────┼───────────┼────────────────────────────
+───────────────┼───────────┼───────────────────────────────────────────────────────────────
           9751 ┊     0.46% ┊ stats_arena_print
-          9503 ┊     0.45% ┊ output
+          9503 ┊     0.45% ┊ std::sys_common::backtrace::output::hf6421f76165dc3d9
```
data-pup: and here's a nice example of what we gain. as the kids say, _"you love to see it."_

parser/object_parse/die_parse/item_name.rs
philipc: This is all subjective, so do what you prefer, but I would do:
```rust
let mut attr = die.attr_value(gimli::DW_AT_linkage_name)?;
if attr.is_none() {
    attr = die.attr_value(gimli::DW_AT_name)?;
}
if let Some(attr) = attr {
    //unchanged
   ...
}
```

data-pup: Thanks for providing some feedback! This led me to think about it a little more, and I think I found a nice way to do this with a `match` statement [here](https://github.com/rustwasm/twiggy/pull/402/files#diff-a5729e8c41bb69cc717e50cab8e2da98R33). No mutability! :tada:

PR #426, opened 2020-02-08
parser/object: don't use Parse trait

philipc:
It complicates things for no benefit.


philipc:
Hide whitespace changes when reviewing.

data-pup:
Thank you!

PR #543, opened 2021-01-29
Upgrade wasmparser dependency

strager:
wasmparser, used by Twiggy, has some bugs which cause Twiggy to crash prematurely. Twiggy is using a very old version of wasmparser. Let's upgrade Twiggy to the latest version of wasmparser.

**Notes for reviewers**:

* Is Twiggy maintained? The latest release was in June 2019, and the latest commit was in July 2020.
* In the patches, I marked code with "@@@" if I was unsure what the code is supposed to do. Treat these comments as TODOs for this PR. I want to do plenty of cleanup before formal code review.
* The patches are a work-in-progress. I'd like high-level feedback before proceeding. (I don't want to spend time on these TODOs if Twiggy is unmaintained.) The patches pass all tests on my machine, but might be missing lots of important test coverage.


axic:
Just to second the poster: it would be nice to have some of these improvements.

AlexEne:
I've published version `0.7` that contains the `wasm32-wasi` fix.
If you want, you could open a PR with your change.
I don't fully uunderstand this either, but I guess it's better to merge if all tests pass now. If it has bugs that have no tests we can fix them as they pop up later and add tests.

AlexEne:
Merged as part of: https://github.com/rustwasm/twiggy/pull/637
Thanks all for the hard work on this.

PR #576, opened 2021-07-16
Skip unknown entries in the custom name section

### LINKED ISSUE #574: Does not work with rustc target wasm32-wasi.
Using: twiggy-opt 0.6.0

----

Code,

```rust
fn main() {
	println!("Hello, world!");
}
```

Cargo.toml,

```toml
[profile.release]
lto = true
opt-level = "s"
```

Build the project,

```text
cargo -v build --release --target wasm32-wasi
   Compiling felipe-wasi v0.1.0 (/home/ecarroll/felipe-wasi)
     Running `rustc --crate-name felipe_wasi --edition=2018 src/main.rs --error-format=json --json=diagnostic-rendered-ansi --crate-type bin --emit=dep-info,link -C opt-level=s -C lto -C metadata=e0e763637e4f02a5 --out-dir /home/ecarroll/felipe-wasi/target/wasm32-wasi/release/deps --target wasm32-wasi -L dependency=/home/ecarroll/felipe-wasi/target/wasm32-wasi/release/deps -L dependency=/home/ecarroll/felipe-wasi/target/release/deps`
    Finished release [optimized] target(s) in 0.99s
```

And then try to run `twiggy`,

```text
$ twiggy top -n 20 ./target/wasm32-wasi/release/felipe-wasi.wasm
error: WASM error: Invalid name type (at offset 276799)
  caused by: Invalid name type (at offset 276799)
```

`wasm-nm` works though

    wasm-nm -z ./target/wasm32-wasi/release/felipe-wasi.wasm

athei:
twiggy uses an ancient version of `wasmparser` which errors out when encountering an "unknown" entry in the custom name section. The error looks like this:

```
error: WASM error: Invalid name type (at offset 276799)
  caused by: Invalid name type (at offset 276799)
```

This PR adds code to just skip any name entry which fails to parse. This is a short term solution to resolve one specific issue which is a blocker for me. I also started to port twiggy to the newest wasmparser version in order to increase compatibility with new wasm features. This is a lot of work, though. The API changed quite a bit.

fixes #574

### DIFF
--- a/parser/wasm_parse/mod.rs
+++ b/parser/wasm_parse/mod.rs
@@ -323,7 +323,14 @@ fn parse_names_section<'a>(
     for IndexedSection(_, section) in indexed_sections.iter() {
         if let wasmparser::SectionCode::Custom { name: "name", .. } = section.code {
             for subsection in section.get_name_section_reader()? {
-                let f = match subsection? {
+                // We use a rather old version of wasmparser. This is a workaround
+                // to skip new types of name subsections instead of aborting.
+                let subsection = if let Ok(subsection) = subsection {
+                    subsection
+                } else {
+                    continue;
+                };
+                let f = match subsection {
                     wasmparser::Name::Function(f) => f,
                     _ => continue,
                 };
@@ -512,7 +519,13 @@ impl<'a> Parse<'a> for wasmparser::NameSectionReader<'a> {
         let mut i = 0;
         while !self.eof() {
             let start = self.original_position();
-            let subsection = self.read()?;
+            // We use a rather old version of wasmparser. This is a workaround
+            // to skip new types of name subsections instead of aborting.
+            let subsection = if let Ok(subsection) = self.read() {
+                subsection
+            } else {
+                continue;
+            };
             let size = (self.original_position() - start) as u32;
             let name = match subsection {
                 wasmparser::Name::Module(_) => "\"module name\" subsection",


HCastano:
Looks like work to update `wasmparser` has already been started in https://github.com/rustwasm/twiggy/pull/543, but it hasn't moved in a while 😞

AlexEne:
This would be quite useful for wasm32-wasi cases. 
Is there any work left that or help needed with in order needed to get this in a ready-to-merge state?

athei:
From my point of view this could be just merged. But It seems that this project is abandoned.

AlexEne:
I'll ask around in the rust-wasm discord. I'd gladly share some maintenance work on this as it's a really helpful tool

AlexEne:
I've tested this locally and existing tests still pass. 
It seems that this doesn't add a new test for the case, but I'll merge as-is for now as I really need this functionality too :D

Before I'll merge anything else, I'll first try to figure out what's up with the CI (maybe github actions may be easier to work with),

Artoria2e5:
Will it be possible to prepare a new git tag? It's a bit annoying to manually tell `cargo` to use git...

AlexEne:
Yes, I can do a release once I finish the current cleanup work. I think updating some of the wasmparser lib will take a while so maybe i'll do one before that.

AlexEne:
I tried publishing a new version, I'm getting a weird error. I will debug this on Tuesday:  
Full logs https://gist.github.com/AlexEne/36311d128ec5ee67ec0c916c01c536e5
```sh
error[E0599]: no variant or associated item named `from_args` found for enum `Options` in the current scope
  --> twiggy.rs:17:33
   |
17 |     let options = opt::Options::from_args();
   |                                 ^^^^^^^^^ variant or associated item not found in `Options`
   |
   = help: items from traits can only be used if the trait is in scope
help: the following trait is implemented but not in scope; perhaps add a `use` for it:
   |
6  | use structopt::StructOpt;
   |

warning: unused import: `structopt::StructOpt`
 --> twiggy.rs:9:5
  |
9 | use structopt::StructOpt;
  |
  = note: `#[warn(unused_imports)]` on by default

error: aborting due to previous error; 1 warning emitted
```

philipc:
The meaning of that error is that two different versions of `structopt` are in use.
You probably need to update these dependency versions:
https://github.com/rustwasm/twiggy/blob/f2e5abb5eeaf614fb5911e748ac40b5ccb836a2d/twiggy/Cargo.toml#L22-L26

AlexEne:
That's strange, both are at 0.3.22 when doing `cargo tree` but maybe `cargo.lock` has different things cached.

philipc:
You're still depending on `twiggy-opt` version 0.6.0 (instead of 0.7.0), which depends on `structopt` 0.2. When building in the workspace, the `path` attribute takes precedence.

AlexEne:
@fitzgen Can I please get crates.io permissions for the other crates so i can publish a new twiggy release? (`twiggy-ir`, etc.)

fitzgen:
Done

AlexEne:
Thank you!

PR #583, opened 2021-08-06
added github actions

AlexEne:
First pass at some github actions.
They just check lints + build.


AlexEne:
These passed on my branch here: https://github.com/AlexEne/twiggy/actions/runs/1106714339

PR #592, opened 2021-08-09
Added guide publishing back

AlexEne:
Added it in github actions. Hopefully this works (I've never tried it before and it's unclear how I could test it - I changed a bit the guide text to confirm it end-to-end).

I also removed some warnings.


AlexEne:
Let's hope this works :)

PR #637, opened 2022-06-03
Upgrade wasmparser and wasm-bindgen

nokotan:
This pull request will take #543 over.

* Upgrade wasmparser from v0.36.0 to v0.73.0
* Upgrade wasm-bindgen from v0.68.0 to v0.80.0


parser/wasm_parse/mod.rs
```diff
@@ -15,50 +15,125 @@ pub struct SectionIndices {
...
+
+    fn eof(&self) -> bool {
+        self.offset == self.data.len()
+    }
+
+    fn read(&mut self) -> Result<wasmparser::Payload<'a>, traits::Error> {
+        let (section, bytes_consumed) =
+            match self.parser.parse(&self.data[self.offset..], self.eof())? {
+                wasmparser::Chunk::NeedMoreData { .. } => {
+                    panic!("@@@ nyi");
```
AlexEne, line 59: Let's return an error here rather than panic since the function is already set up for that

parser/wasm_parse/mod.rs
```diff
@@ -15,50 +15,125 @@ pub struct SectionIndices {
...
-                wasmparser::SectionCode::Code => code_section = Some(indexed_section),
-                wasmparser::SectionCode::Function => function_section = Some(indexed_section),
+            let at_eof = self.offset == self.data.len();
+            if at_eof {
+                break;
+            }
+            let (section, bytes_consumed) =
+                match self.parser.parse(&self.data[self.offset..], at_eof)? {
+                    wasmparser::Chunk::NeedMoreData { .. } => {
+                        panic!("@@@ nyi");
```
AlexEne, line 110: We should return an error here too.

parser/wasm_parse/mod.rs
```diff
@@ -15,50 +15,125 @@ pub struct SectionIndices {
...
+                    }
+                    wasmparser::Chunk::Parsed { consumed, payload } => (payload, consumed),
+                };
+            self.offset += bytes_consumed;
+            let size = self.current_position() - start; // @@@ refactor this nonsense
```
AlexEne, line 115: Forgot this comment in?

parser/wasm_parse/mod.rs
```diff
@@ -15,50 +15,125 @@ pub struct SectionIndices {
...
+                }
+                wasmparser::Payload::CodeSectionEntry { .. } => {
+                    // Ignore.
+                }
                 _ => sections.push(indexed_section),
             };
-            sizes.insert(idx, size as u32);
-            idx += 1;
+            sizes.insert(idx, size as u32); // @@@ delete if unused.
+            idx += 1; // @@@ wrong. we're counting any payload as a section.
```
AlexEne, line 134: Should this be adjusted?

parser/wasm_parse/mod.rs
```diff
@@ -195,42 +284,45 @@ impl<'a> Parse<'a> for wasmparser::ModuleReader<'a> {
                             wasmparser::ImportSectionEntryType::Global(_) => {
                                 indices.globals.push(id);
                             }
+                            wasmparser::ImportSectionEntryType::Event(_) => {} // @@@
```
AlexEne, line 287: Now it seems to only count functions and globals and that's fine. Let's just remove the comments tho.

parser/wasm_parse/mod.rs
```diff
@@ -242,77 +334,97 @@ impl<'a> Parse<'a> for wasmparser::ModuleReader<'a> {
...
+                    }
+                    .parse_edges(items, ())?;
                 }
-                wasmparser::SectionCode::Code | wasmparser::SectionCode::Function => {
+                wasmparser::Payload::CodeSectionStart { .. }
+                | wasmparser::Payload::FunctionSection { .. } => {
                     unreachable!("unexpected code or function section found");
                 }
+
+                _ => {} // @@@
```
AlexEne, line 392: Let's replace this with the explicit things that we have left that are unhandled. Should it also be ignored, or point to the unreachable above?

parser/wasm_parse/mod.rs
```diff
@@ -242,77 +334,97 @@ impl<'a> Parse<'a> for wasmparser::ModuleReader<'a> {
...
+        wasmparser::Payload::CodeSectionStart { .. } => get_code_section_name(),
+        wasmparser::Payload::DataSection(_) => "data section headers".to_string(),
+        wasmparser::Payload::DataCountSection { .. } => "data count section headers".to_string(),
+        wasmparser::Payload::Version { .. } => "wasm magic bytes".to_string(),
+
+        wasmparser::Payload::CodeSectionEntry { .. } => {
+            panic!("unexpected CodeSectionEntry");
+        }
+
+        _ => format!("@@@ {:?}", section), // @@@
```
AlexEne, line 427: Same here, let's just add all the remaining sections explicitly vs using `_`.

parser/wasm_parse/mod.rs
```diff
@@ -404,18 +517,24 @@ impl<'a> Parse<'a> for (IndexedSection<'a>, IndexedSection<'a>) {
...
         let added = items.size_added() - start;
-        let size = sizes
-            .get(&code_section_idx)
-            .ok_or_else(|| traits::Error::with_msg("Could not find section size"))?
-            + sizes
-                .get(&func_section_idx)
-                .ok_or_else(|| traits::Error::with_msg("Could not find section size"))?;
+        let code_section_size = code_section.byte_size as u32;
+        let func_section_size = func_section.byte_size as u32;
+        // @@@ bye
```
AlexEne, line 528: This looks ok to me, but we probably want to remove the commented code.

parser/wasm_parse/mod.rs
```diff
@@ -589,31 +707,38 @@ impl<'a> Parse<'a> for wasmparser::TypeSectionReader<'a> {
...
+                                name.push_str(ty2str(*result));
+                            }
+                            name.push_str(")");
+                        }
                     }
-                    name.push_str(")");
+
+                    items.add_item(ir::Item::new(id, name, size, ir::Misc::new()));
                 }
+                // @@@ Is this correct?
```
AlexEne, line 738: To be honest I don't know if this is correct, but it looks like it preserves the previous behavior so i guess yes :D

parser/wasm_parse/mod.rs
```diff
@@ -636,7 +761,7 @@ impl<'a> Parse<'a> for wasmparser::ImportSectionReader<'a> {
         for (i, imp) in iterate_with_size(self).enumerate() {
             let (imp, size) = imp?;
             let id = Id::entry(idx, i);
-            let name = format!("import {}::{}", imp.module, imp.field);
+            let name = format!("import {}::{}", imp.module, imp.field.unwrap_or("@@@"));
```
AlexEne, line 764: Let's use `unwrap_or("unknown")`?

parser/wasm_parse/mod.rs
```diff
@@ -762,14 +887,21 @@ impl<'a> Parse<'a> for wasmparser::ExportSectionReader<'a> {
                 wasmparser::ExternalKind::Global => {
                     items.add_edge(exp_id, indices.globals[exp.index as usize]);
                 }
+                wasmparser::ExternalKind::Event
+                | wasmparser::ExternalKind::Type
+                | wasmparser::ExternalKind::Module
+                | wasmparser::ExternalKind::Instance => {} // @@@
```
AlexEne, line 893: I don't think there's a need for the comment here

parser/wasm_parse/mod.rs
```diff
@@ -931,8 +1071,9 @@ fn ty2str(t: Type) -> &'static str {
         Type::F32 => "f32",
         Type::F64 => "f64",
         Type::V128 => "v128",
-        Type::AnyFunc => "anyfunc",
-        Type::AnyRef => "anyref",
+        Type::ExnRef => "exnref",    // @@@
```
AlexEne, line 1074: Can we rename this since it's a bit strange for externRef to be anyref?

twiggy/tests/all/main.rs
```diff
@@ -53,6 +53,7 @@ macro_rules! test {
 
             let actual = String::from_utf8_lossy(&output.stdout);
 
+            // @@@ unsure about the test changes I introduced in this patch. =S
```
AlexEne, line 56: The only one I'm not sure about is `wee_alloc_csv` test. The other ones seem fine.
I'm just making a note here to double check those numbers are as expected.
I think the comment can be safely deleted.

AlexEne (commented):
Thank you so much for the PR.
I've added a bunch of comments, nothing major.

AlexEne (approved):

AlexEne:
Thanks for your hard work. I still have a bit of questions about the diff in tests but that's ok.
I'll merge this and if there are any bugs we can fix them before the next release, no need to keep postponing this until it's fully 100% "ready" and "correct" :D.

PR #699, opened 2023-03-30
Support data segment name

kateinoigakukun:
- Upgrade wasmparser from 0.73.0 to 0.80.0, which supports data subsection in the name section.
- Add support for data segment name


AlexEne:
Thank you so much for this contribution!

PR #477, opened 2020-06-24
(WIP) Parse Threaded Modules

data-pup:
### Parse Threaded Modules

:construction: WIP :construction: 

Fixes #326.

This continues the work from #365.

First, a reduced test fixture based on the repro posted in the original issue [here](https://github.com/rustwasm/twiggy/issues/326#issuecomment-552558805):

```
;; `threads.wat` fixture
;;
;; Built with `wat2wasm --enable-bulk-memory --enable-threads`
(module
 (memory $0 1 1)
 (data (i32.const 0) "")
 (func $f (data.drop 0))
)
```

After building the `.wasm` file with `wat2wasm --enable-bulk-memory --enable-threads`, we have a `44` byte file, with the following header sizes: 

```
; exa -l --header twiggy/tests/all/fixtures/threads.wasm

Permissions Size User    Date Modified Name
.rw-rw-r--    44 katelyn 21 Jun 23:07  twiggy/tests/all/fixtures/threads.wasm


; wasm-objdump --headers twiggy/tests/all/fixtures/threads.wasm

threads.wasm:	file format wasm 0x1

Sections:

     Type start=0x0000000a end=0x0000000e (size=0x00000004) count: 1
 Function start=0x00000010 end=0x00000012 (size=0x00000002) count: 1
   Memory start=0x00000014 end=0x00000018 (size=0x00000004) count: 1
DataCount start=0x0000001a end=0x0000001b (size=0x00000001) count: 1
     Code start=0x0000001d end=0x00000024 (size=0x00000007) count: 1
     Data start=0x00000026 end=0x0000002c (size=0x00000006) count: 1
```

And we can print sizes without a `should not parse the same key into multiple items` panic!

```
; target/debug/twiggy top twiggy/tests/all/fixtures/threads.wasm

 Shallow Bytes │ Shallow % │ Item
───────────────┼───────────┼───────────────────────
             8 ┊    18.18% ┊ wasm magic bytes
             7 ┊    15.91% ┊ code[0]
             6 ┊    13.64% ┊ code section headers
             5 ┊    11.36% ┊ data[0]
             3 ┊     6.82% ┊ type[0]: () -> nil
             3 ┊     6.82% ┊ type section headers
             3 ┊     6.82% ┊ memory[0]
             3 ┊     6.82% ┊ memory section headers
             3 ┊     6.82% ┊ "data count" section
             3 ┊     6.82% ┊ data section headers
            44 ┊   100.00% ┊ Σ [10 Total Rows]
```

[...]

PR #744, opened 2023-11-29
Do not duplicate `DataCountSection` and `StartSection` root items

Maaarcocr:
# why?

it should fix https://github.com/rustwasm/twiggy/issues/326

# how? 

I've taken some test code from https://github.com/rustwasm/twiggy/pull/477 but decided to try to figure out how it was possible that an ID was repeated more than once. It turns out that for the `DataCountSection` and `StartSection` we would end up adding 2 root items that have the exact same id. I just deleted the code that did this. 

I think this is now correct, but I may be wrong.


laurentmmeyer:
For me it fixed the panic crash that I had with one emscripten binary, can we have a new release? seems to work :)

PR #755, opened 2025-03-31
Much better support for ELF files

bjorn3:
This now uses symbols rather than debuginfo as source of truth of which functions exist. This allows analyzing binaries which don't have debuginfo and ensures symbols for which rustc doesn't generate debuginfo for whatever reason are still accounted for.

For ELF files we now take the runtime allocated size rather than the file size into account. When developing for microcontrollers the runtime allocated size is what matters, not the file size as only the runtime allocated segments actually get flashed onto the device. The rest of the ELF file is ignored.

This also adds support for getting the graph of references between symbols. This is done asking the user to compile with `--emit-relocs` and then reading the resulting relocations. Unlike the twiggy backend for WASM this supports both references originating from functions as well as references originating from data objects. So it will handle vtables and references in statics just fine.

This contains three commits regenerating hello_elf to see step by step how the command outputs improve by these changes. The regenerations should probably be squashed together into a single commit before merging to save a couple of MB in repo size.


bjorn3:
Yeah, locally tests pass.

AlexEne:
Thanks for this. I also added you to maintainers so you don't have to wait on me for other PRs

bjorn3:
Fixed CI in https://github.com/rustwasm/twiggy/pull/756.

PR #756, opened 2025-03-31
Fix CI

bjorn3:


bjorn3:
This PR fixes CI. In the future CI should move away from actions-rs as those are now archived.

AlexEne:
Hi, thanks for all the good work you're doing here. I added you to crates.io owners if you want to publish a new version :)

PR #769, opened 2025-05-10
Update wasmparser to v0.230.0, adjust parser to support this

0xJonas:
This does not add any meaningful new features to the parser, but only serves to enable parsing WASM modules which make use of more recent proposals to the WASM spec.


AlexEne:
Amazing work, thanks for doing this!

AlexEne (approved):

